PR #4, opened 2020-09-19
Fix code overflow without highlightjs

### LINKED ISSUE #3: Code blocks overflow without highlightjs
Before highlightjs loading, code blocks are overflowing on the right, without possibility to scroll horizontally.
overflow

This make the overflowing code appear and then disappear during loading.

And for visitors using NoScript, it stays overflowing and can be partly unreadable if it goes beyond screen limits.

Adding `overflow-x: auto;` to the CSS fixes it and allows to scroll horizontally.
fixed

I have a pull request on the way.

pyvain:
Fixes #3

### DIFF
--- a/assets/css/post-single.css
+++ b/assets/css/post-single.css
@@ -211,6 +211,7 @@
     color: rgba(255, 255, 255, 0.8);
     background: transparent;
     border-radius: 0;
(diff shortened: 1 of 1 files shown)


adityatelange:
Thank You for contributing : )

PR #6, opened 2020-09-19
Fix interversion of next and prev translations in Spanish

pyvain:
Spanish translations for "next" and "previous" were swapped.


adityatelange:
Merged, thanks!

PR #10, opened 2020-09-22
Update zh.yaml

EV7:


i18n/zh.yaml
```diff
@@ -3,3 +3,14 @@
...
+- id: read_time
+  translation:
+    one : "1 分钟"
+    other: "{{ .Count }} 分钟"
+
+- id: toc
+  translation: "目录"
+
+- id: translations
+  translation: "语言"
```
adityatelange, line 16: are you sure about this translation ? It should mean "Translations" .
On Google Translate it shows languages though...

EV7: I don't know how to explain the difference precisely.
These two words are similar in this context, **both can be accepted**.

The word "translation" can be directly translated to "翻译", which implicitly hints this article is translated by someone other than the author (especially a machine in this context).

In the other hand, it can be translated to "语言" too, which doesn't hint anything, it's a neutral word.

I see your `single.html#18`, and you `range .Translations`, if I don't misunderstand, this statement ranges “xxx.en.md”, "xxx.zh.md", "xxx.fr.md"..... 

You can't say these multi-language version articles must be translated by others, so I think that use a neutral word "语言" is better.

Well... language is complex, I have to give you a huge paragraph to explain.......

adityatelange: okay great ... thanks for conforming

PR #12, opened 2020-10-01
Added RSS feathers icon and provide per page comments toggle

TheClonker:
Hey,

two small changes, adding a RSS icon and provide a new page config option.
Regarding the comment toggle, i'm not sure if that is best practice or if something could go wrong if the param is not defined etc.


adityatelange:
It is always a good practise to open different pull requests for unrelated changes. Like in this case I wish to merge https://github.com/adityatelange/hugo-PaperMod/pull/12/commits/daef09a904a73220295d39d65f7ab22292f9b61d but may want some review on https://github.com/adityatelange/hugo-PaperMod/pull/12/commits/0867691c3163d08589235aeeeff982edf4c66954.
That makes merging complicated.

TheClonker:
Yeah you are right, i thought since it's such a small change i didn't wanna spam PR's.
I mean daef09a is just extending and no bug fix etc but i get your point.

adityatelange:
Anyways, I just tested it and it is how it should be  :+1:  , merging
Thank you for contributing :D

PR #33, opened 2020-10-11
hide empty date

tomkukral:
Don't show post time instead of zero date.


adityatelange:
@tomkukral What are the cases this might be helpful ?
Like.. why do you want to hide only date and not other meta about the post ?

This also adds an leading `•` : separator which looks weird.

PR #39, opened 2020-10-16
Add DateFormat theme parameter

Northerner-coder:
Hi! Thanks for great theme!  

I think allowing user to set date format throughout the site and not be limited to "January 2, 2006"-format might come in handy, at least in my case.  I added handling of .Site.Params.DateFormat variable to the templates.   If parameter is not set it defaults to usual "January 2, 2006" so current users should not be affected.

Please let me know what you think.


adityatelange:
Works Fine :+1: 
Thank You for contributing.

PR #43, opened 2020-10-17
Responsive cover images

Kartoffel:
I am working on adding responsive sizing for cover images (using _srcset_ to provide multiple sizes of the cover image, similar to [this](https://laurakalbag.com/processing-responsive-images-with-hugo/)).

The first step is to move the cover image to a new partial, as this was shared between the homepage and single posts.

### DIFF
--- a/layouts/partials/cover.html
+++ b/layouts/partials/cover.html
@@ -4,9 +4,10 @@
 {{- if .Params.cover.image }}
 
 {{- $sizes := (slice "360" "480" "720" "1080" "1500") }}
+{{- $processableFormats := (slice "jpg" "png" "tif" "bmp" "gif") }}
 
 {{- $file := (.Page.Resources.ByType "image").GetMatch (printf "*%s*" (.Params.cover.image)) }}
-{{- if not (isset $file "Width") }}{{ $file = false }}{{ end }} {{/* Exception for non-supported images */}}
+{{- with $file }} {{ if not (in $processableFormats .MediaType.SubType) }} {{ $file = false }} {{ end }}{{ end }}
 
 {{- $src := cond (default true .Site.Params.responsiveImages) $file false }}
 {{- $fallback := cond (default false .Params.cover.relative) (path.Join .RelPermalink .Params.cover.image) (.Params.cover.image) -}}
--- a/layouts/partials/cover.html
+++ b/layouts/partials/cover.html
@@ -16,15 +16,15 @@
 
 <figure class="entry-cover">  
 {{- if $responsiveImg -}}
-    {{- if $addLink }}<a href="{{ $src.RelPermalink | absURL }}" target="_blank" 
+    {{- if $addLink }}<a href="{{ $src.Permalink }}" target="_blank" 
         {{- with .Params.cover.caption }}title="{{ . | plainify }}"{{ else }}{{ with .Params.cover.alt }}title="{{ . | plainify }}"{{ end }}{{ end }}>
     {{- end }}
     <img sizes="(min-width: 768px) 720px, 100vw"
     srcset='{{- range $sizes -}}
-        {{ if ge $src.Width . }}{{ ($src.Resize (printf "%sx" .)).RelPermalink | absURL }} {{ (printf "%sw" .) }},{{ end }}
-    {{- end -}} {{$src.RelPermalink | absURL }} {{printf "%dw" ($src.Width)}}'
-    {{- if ge $src.Width "720" }}src="{{ ($src.Resize "720x").RelPermalink | absURL }}"
-    {{- else }}src="{{ $src.RelPermalink | absURL }}" {{ end }}{{ with .Params.cover.alt }} alt="{{ . | plainify }}"{{ end }}>
+        {{ if ge $src.Width . }}{{ ($src.Resize (printf "%sx" .)).Permalink }} {{ (printf "%sw" .) }},{{ end }}
+    {{- end -}} {{$src.Permalink }} {{printf "%dw" ($src.Width)}}'
+    {{- if ge $src.Width "720" }}src="{{ ($src.Resize "720x").Permalink }}"
+    {{- else }}src="{{ $src.Permalink }}" {{ end }}{{ with .Params.cover.alt }} alt="{{ . | plainify }}"{{ end }}>
     
 {{- else }}{{/* Fall back to full image */}}
     {{- if $addLink }}<a href="{{ $fallback | absURL }}" target="_blank" 
--- a/layouts/partials/cover.html
+++ b/layouts/partials/cover.html
@@ -3,7 +3,7 @@
 {{- with .cxt}} {{/* Apply proper context from dict */}}
 {{- if .Params.cover.image }}
 
-{{- $sizes := (slice "480" "720" "1080" "1500") }}
+{{- $sizes := (slice "360" "480" "720" "1080" "1500") }}
 
 {{- $src := cond (default true .Site.Params.responsiveImages) (.Page.Resources.GetMatch (printf "*%s*" (.Params.cover.image))) false }}
 {{- $fallback := cond (default false .Params.cover.relative) (path.Join .RelPermalink .Params.cover.image) (.Params.cover.image) -}}
--- a/layouts/partials/cover.html
+++ b/layouts/partials/cover.html
@@ -5,13 +5,14 @@
 
 {{- $sizes := (slice "480" "720" "1080" "1500") }}
 
-{{- $src := .Page.Resources.GetMatch (printf "*%s*" (.Params.cover.image)) }}
-{{- $fallback := cond .Params.cover.relative (path.Join .RelPermalink .Params.cover.image) (.Params.cover.image) -}}
+{{- $src := cond (default true .Site.Params.responsiveImages) (.Page.Resources.GetMatch (printf "*%s*" (.Params.cover.image))) false }}
+{{- $fallback := cond (default false .Params.cover.relative) (path.Join .RelPermalink .Params.cover.image) (.Params.cover.image) -}}
 
 {{- $addLink := and (not $IsHome) (default true .Site.Params.linkFullImages) -}}
+{{- $responsiveImg := and $src (default false .Params.cover.relative) -}}
 
 <figure class="entry-cover">  
-{{- if and $src .Params.cover.relative -}}
+{{- if $responsiveImg -}}
     {{- if $addLink }}<a href="{{ $src.RelPermalink }}" target="_blank" title="Original image">{{ end }}
     <img sizes="(min-width: 768px) 720px, 100vw"
     srcset='{{- range $sizes -}}
--- a/layouts/partials/cover.html
+++ b/layouts/partials/cover.html
@@ -5,7 +5,10 @@
 
 {{- $sizes := (slice "360" "480" "720" "1080" "1500") }}
 
-{{- $src := cond (default true .Site.Params.responsiveImages) (.Page.Resources.GetMatch (printf "*%s*" (.Params.cover.image))) false }}
+{{- $file := (.Page.Resources.ByType "image").GetMatch (printf "*%s*" (.Params.cover.image)) }}
+{{- if not (isset $file "Width") }}{{ $file = false }}{{ end }} {{/* Exception for non-supported images */}}
+
+{{- $src := cond (default true .Site.Params.responsiveImages) $file false }}
 {{- $fallback := cond (default false .Params.cover.relative) (path.Join .RelPermalink .Params.cover.image) (.Params.cover.image) -}}
 
 {{- $addLink := and (not $IsHome) (default true .Site.Params.linkFullImages) -}}
--- a/layouts/partials/cover.html
+++ b/layouts/partials/cover.html
@@ -13,7 +13,9 @@
 
 <figure class="entry-cover">  
 {{- if $responsiveImg -}}
-    {{- if $addLink }}<a href="{{ $src.RelPermalink | absURL }}" target="_blank" title="Original image">{{ end }}
+    {{- if $addLink }}<a href="{{ $src.RelPermalink | absURL }}" target="_blank" 
+        {{- with .Params.cover.caption }}title="{{ . | plainify }}"{{ else }}{{ with .Params.cover.alt }}title="{{ . | plainify }}"{{ end }}{{ end }}>
+    {{- end }}
     <img sizes="(min-width: 768px) 720px, 100vw"
     srcset='{{- range $sizes -}}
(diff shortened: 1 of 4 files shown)


adityatelange:
Shouldn't the whole Feature be in one pull request ?
You can add commits to a branch of your forked repo, and open up only one pull request for it :)
refer : https://github.com/adityatelange/hugo-PaperMod/pull/36

---

Also talking about the feature you are implementing, can you explain me more about it ? 

I see that https://laurakalbag.com/processing-responsive-images-with-hugo/ here they are using a handy short-code... so why not just add the short-code and make it compatible to to the theme (if it is not already), rather than natively implementing it in ?

Kartoffel:
Sure, I'll explain a little bit more.

You'll no doubt have read the explanation for using the HTML5 `srcset` parameter, it's very useful for speeding up the website on slow connections as well as providing optional high resolution versions for HiDPI displays.
The shortcode is great for blog content, [one has even been created as a drop-in replacement for hugo's `figure` shortcode](https://cpbotha.net/2020/05/02/drop-in-replacement-for-hugo-figure-shortcode-with-responsive-img-srcset/). Shortcodes however do not work for the cover images as they are defined through the theme, hence I am working to implement it directly in the theme.

It's a bit silly to submit a single PR for moving the cover image to a new partial first, that's mostly for my comfort as I don't tend to develop as quickly as this repo is moving and it will cause fewer merge conflicts later on.
I'll continue working on the responsive cover images, expect some more commits under this PR a bit later :)

singhkays:
> Shortcodes however do not work for the cover images as they are defined through the theme, hence I am working to implement it directly in the theme.

+1. I'm using srcset on my site to avoid loading large files on LowDPI displays but the cover image takes a while to load because it's a large file meant to look good on HiDPI displays. This would be very useful!

Kartoffel:
@singhkays look no further, I've just implemented responsive cover images.

By default it generates 480, 720, 1080, and 1500px alternatives and when you click on the image you get taken to the original size. If you'd prefer different sizes, edit `$sizes` in `layouts/partials/cover.html`.

@adityatelange what do you think?

adityatelange:
> @singhkays look no further, I've just implemented responsive cover images.
> 
> By default it generates 480, 720, 1080, and 1500px alternatives and when you click on the image you get taken to the original size. If you'd prefer different sizes, edit `$sizes` in `layouts/partials/cover.html`.
> 
> @adityatelange what do you think?

Haven't looked at the code yet, but as as per your explanation this is a neat feature. 
If its gonna increase performance I'm in :+1:

Kartoffel:
> If its gonna increase performance I'm in

It'll for sure improve loading speeds for slower network connections, and allow for nice sharp images on HiDPI screens at the same time :)

Here's a (cleaned up) example of what `img` tags it generates: 
``

adityatelange:
@Kartoffel Ping me after you've pushed all necessary changes, and it is ready to be merged :)

Kartoffel:
All ready now! I've added a site parameter to choose whether to include a hyperlink to the full-size image when responsive cover images are used.

Updated the documentation as well, see #53.

Kartoffel:
Totally done for now, pinky promise!

adityatelange:
@Kartoffel 

1. It works only on `relative` covers and not on absolute cover ( leaving external links) ?
2. `cover.relative` needs to be explicitly set. [this](https://github.com/adityatelange/hugo-PaperMod/pull/43/files#diff-dd30a18c966d1e96b055d5c9955d9d8ef1200159705fb5a6323ceb4a89db5be7R9) won't work if  `cover.relative` is not present.
3. We need an option to _disable_ responsive cover images. (For many pages with large images it will take time to generate those). Also not everyone needs responsive images.

Kartoffel:
@adityatelange 
> 1. It works only on `relative` covers and not on absolute cover ( leaving external links) ?
> 2. `cover.relative` needs to be explicitly set. [this](https://github.com/adityatelange/hugo-PaperMod/pull/43/files#diff-dd30a18c966d1e96b055d5c9955d9d8ef1200159705fb5a6323ceb4a89db5be7R9) won't work if  `cover.relative` is not present.
> 3. We need an option to _disable_ responsive cover images. (For many pages with large images it will take time to generate those). Also not everyone needs responsive images.

1. Correct, and additionally the relative image needs to be in the [Page Bundle](https://gohugo.io/content-management/page-bundles/) for this to work, otherwise `.Page.Resources` can't find the file. External links and relative images that can't be found will be added in the old way ([this part](https://github.com/adityatelange/hugo-PaperMod/pull/43/files#diff-dd30a18c966d1e96b055d5c9955d9d8ef1200159705fb5a6323ceb4a89db5be7R23)).
2. You're right, I failed to notice that. What should the default be when `cover.relative` is not present? Currently it's an absolute URL by default.
3. You're right, I'll add the option. Responsive images by default or only optionally?

Thanks for the feedback!

Kartoffel:
@adityatelange fixed the error on unset `relative` parameter, defaults are now absolute URLs and responsive images enabled.

There is a new site parameter to globally enable/disable the responsive images, I've also added this to the documentation site in #53.

adityatelange:
> @adityatelange
> Fixed the error on unset `relative` parameter, defaults are now absolute URLs and responsive images enabled.
> 
> There is a new site parameter to globally enable/disable the responsive images, I've also added this to the documentation site in #53.

Nice !

layouts/partials/cover.html
```diff
@@ -0,0 +1,35 @@
+{{- $IsHome := .IsHome }}
+
+{{- with .cxt}} {{/* Apply proper context from dict */}}
+{{- if .Params.cover.image }}
+
+{{- $sizes := (slice "480" "720" "1080" "1500") }}
```
adityatelange, line 6: Can we add `360` here

layouts/partials/cover.html
```diff
@@ -0,0 +1,35 @@
...
+
+{{- $src := cond (default true .Site.Params.responsiveImages) (.Page.Resources.GetMatch (printf "*%s*" (.Params.cover.image))) false }}
+{{- $fallback := cond (default false .Params.cover.relative) (path.Join .RelPermalink .Params.cover.image) (.Params.cover.image) -}}
+
+{{- $addLink := and (not $IsHome) (default true .Site.Params.linkFullImages) -}}
+{{- $responsiveImg := and $src (default false .Params.cover.relative) -}}
+
+<figure class="entry-cover">  
+{{- if $responsiveImg -}}
+    {{- if $addLink }}<a href="{{ $src.RelPermalink }}" target="_blank" title="Original image">{{ end }}
```
adityatelange, line 16: We should add `| absURL` to every `.RelPermalink`.

adityatelange (commented):
@Kartoffel 

1. I see that newer image formats like "webp" which google recommends isn't supported by this. 
A filter should be added.
2. If the resource is not an image, the code tries to resize it ( this is also same when linked URI is not present)
  giving error as `resources.genericResource is not an image`

layouts/partials/cover.html
Kartoffel: Of course!

layouts/partials/cover.html
```diff
@@ -0,0 +1,35 @@
...
+    <img sizes="(min-width: 768px) 720px, 100vw"
+    srcset='{{- range $sizes -}}
+        {{ if ge $src.Width . }}{{ ($src.Resize (printf "%sx" .)).RelPermalink }} {{ (printf "%sw" .) }},{{ end }}
+    {{- end -}} {{$src.RelPermalink}} {{printf "%dw" ($src.Width)}}'
+    {{- if ge $src.Width "720" }}src="{{ ($src.Resize "720x").RelPermalink }}"
+    {{- else }}src="{{ $src.RelPermalink }}" {{ end }}{{ with .Params.cover.alt }} alt="{{ . | plainify }}"{{ end }}>
+    
+{{- else }}{{/* Fall back to full image */}}
+    {{- if $addLink }}<a href="{{ $fallback | absURL }}" target="_blank" title="Original image">{{ end }}
```
adityatelange, line 25: Let the `title` be `caption` else `alt`.
"Original image" won't look good for multilingual content

layouts/partials/cover.html
Kartoffel: I'm not sure why you insist on absolute permalinks, but I've added them.

layouts/partials/cover.html
Kartoffel: Done.

Kartoffel:
@adityatelange 
> 1. I see that newer image formats like "webp" which google recommends isn't supported by this.
>    A filter should be added.
> 2. If the resource is not an image, the code tries to resize it ( this is also same when linked URI is not present)
>    giving error as `resources.genericResource is not an image`

Both of these are fixed now. Only `Image` type page resources are matched, and after that a small check is placed to see if they are actually processable by Hugo (to avoid the error on "webp"-type images).

Please let me know if you have any other comments.

adityatelange (commented):
@Kartoffel 
```bash
WARNING: calling IsSet with unsupported type "invalid" (<nil>) will always return false.

WARNING: calling IsSet with unsupported type "ptr" (*resources.resourceAdapter) will always return false.
```
The Relative images are not processes now, directly links to the og image.

Kartoffel:
@adityatelange Dang, I should've spotted that.

Can you find another way to do effectively the same? I'd hate to have to hardcode the image types supported for resizing.

Kartoffel:
@adityatelange there's the simple check, this should work until Hugo has another way to test if an image can be processed.

adityatelange:
@Kartoffel I have re-written code with help of your previous work, but I am unaware how to push to a pull-request.
How should we proceed ?

changes are
- reduced some code lines
- added jpeg to file format, for some .jpg files hugo detects them as jpeg
- Site config vars are inside cover param
```yml
params:
    cover:
      responsiveImages: true
      linkFullImages: false
```

Kartoffel:
@adityatelange 

That sounds great!

I have enabled maintainer edits to this pull request, so you should be able to add commits directly to this pull request following [this tutorial](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/committing-changes-to-a-pull-request-branch-created-from-a-fork).

adityatelange:
@Kartoffel 
Can you check whether everything works well ?
I feel its working fine but still 2nd person code review is always better :)

Kartoffel:
@adityatelange nice work :)

This gives a `at <$cover.MediaType.SubType>: nil pointer evaluating resource.Resource.MediaType` when `cover.relative` is `true` and the image doesn't exist in the page bundle.

Other than that, all seems good!

adityatelange:
> @adityatelange nice work :)
> 
> This gives a `at <$cover.MediaType.SubType>: nil pointer evaluating resource.Resource.MediaType` error when `cover.relative` is `true` and the image doesn't exist in the page bundle.
> 
> Other than that, all seems good!

@Kartoffel What I have observed is :
if the image is in pageBundle the image is relative for sure ! (please correct me if I am wrong)
so we need not have the `relative=true` param at all, we can just check if the cover exists in page bundle :)

I'll push more reduced code soon

Kartoffel:
@adityatelange

> What I have observed is :
> if the image is in pageBundle the image is relative for sure ! (please correct me if I am wrong)
> so we need not have the `relative=true` param at all, we can just check if the cover exists in page bundle :)

I believe so! I made that observation earlier as well, but thought that you might've left the `relative` parameter in for another purpose.

Excellent improvement, I'd say we've arrived at the most elegant solution yet!

I've spent some time testing different cases and have not been able to break this one.  It should be ready to merge :tada:

adityatelange:
@Kartoffel 
> I believe so! I made that observation earlier as well, but thought that you might've left the `relative` parameter in for another purpose.

That is true, it is being used in templates, have to make proper arrangements for it then merge it.

Kartoffel:
@adityatelange 
> or we could merge it and will work same as before, no one would get affected !

I don't see any issue with this, the code should be compatible with all previous usage and I can't find any 'standardisation' of a `relative` parameter in other templates.

Kartoffel:
Wonderful! I'm really excited that this feature is now available for every user of the theme.

Thanks for sticking with me, it was a while since I'd worked with Hugo templates and it's been a great learning experience :smile:

adityatelange:
> Wonderful! I'm really excited that this feature is now available for every user of the theme.
> 
> Thanks for sticking with me, it was a while since I'd worked with Hugo templates and it's been a great learning experience :smile: 

Yes, same pinch!

Good Job @Kartoffel :tada:
Thank you for contributing.. :)

adityatelange:
@Kartoffel I'm thinking about only enabling it in production environment, what do you think ?

Kartoffel:
@adityatelange that's an option, but rather than fully disabling them in development I'd leave maybe just one other size next to the original. It will speed things up, but still show the responsive images working.

adityatelange:
> @adityatelange that's an option, but rather than fully disabling them in development I'd leave maybe just one other size next to the original. It will speed things up, but still show the responsive images working.

Already pushed.. 
Anyways, `params.env = production` can get it working, if needed

Kartoffel:
@adityatelange it took me a while, but I found a little issue stemming from this feature

We've removed the use of `cover.relative`, but this is still used in some of the templates including `twitter_cards.html`. This means that you get incorrect image URLs like

`<meta property="og:image" content="http://example.com/featured.jpg" />`

instead of

`<meta property="og:image" content="http://example.com/2021/post-name/featured.jpg" />`

i.e. the cover image is incorrectly taken as absolute URL when it should be relative.

Do you think that the relative parameter should be added back, or can we check if the image is present in the page bundle like the `partials/cover.html` does?

adityatelange:
> @adityatelange it took me a while, but I found a little issue stemming from this feature
> 
> We've removed the use of `cover.relative`, but this is still used in some of the templates including `twitter_cards.html`. This means that you get incorrect image URLs like
> 
> `<meta property="og:image" content="http://example.com/featured.jpg" />`
> 
> instead of
> 
> `<meta property="og:image" content="http://example.com/2021/post-name/featured.jpg" />`
> 
> i.e. the cover image is incorrectly taken as absolute URL when it should be relative.
> 
> Do you think that the relative parameter should be added back, or can we check if the image is present in the page bundle like the `partials/cover.html` does?

We haven't removed `cover.relative`. We are just not using it with cover's code as it is works fine without having a check on it . 

That means it is necessary to use `cover.relative` to work well everywhere but it is not needed to check for it within cover's partial. ;)

Kartoffel:
That's another way to phrase it, but fair enough. In fact, I've already updated all my posts to include the `cover.relative` parameter and that fixed it.

I can't seem to find the parameter in the documentations, is it still there?

adityatelange:
> I can't seem to find the parameter in the documentations, is it still there?

Yes it's not, a lot of documentation is incomplete, I am just being lazy to do all the documentation lol

mukhtharcm:
Is there any way to use these responsive images without page bundle?

PR #44, opened 2020-10-17
Add placeholder partial to baseof.html to be overwritten outside of the theme

Northerner-coder:
Hi! I was struggling with injecting custom code into <head> on every page to enable analog of Google Analytics on the blog. The only way to do it in Hugo I found is to overwrite head.html or baseof.html partial from outside of the theme.  However, it would require copying the code of the partial and updating it appropriately as the theme updates.

Another solution is to add to the theme a placeholder partial without any useful content into the baseof.html which in turn can be overwritten from outside of the theme. The same approach is already used in case of comments.html.

Please let me know if there is an easier way )

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -3,6 +3,7 @@
 
 <head>
     {{- partial "head.html" . }}
+    {{- partial "head_custom_content.html" -}}
 </head>
 
 <body class="{{ if (and (eq .Kind `page` ) (ne .Layout `archives`)) }}single{{ else }}list
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -3,7 +3,6 @@
 
 <head>
     {{- partial "head.html" . }}
-    {{- partial "head_custom_content.html" -}}
 </head>
 
 <body class="{{ if (and (eq .Kind `page` ) (ne .Layout `archives`)) }}single{{ else }}list
--- a/layouts/partials/extend_head.html
+++ b/layouts/partials/extend_head.html
@@ -1,0 +1,5 @@
+<!-- Head custom content area start -->
+<!--
+Insert any custom code (web-analytics, resources, etc.) - it will appear in the <head></head> section of every page. Can be overwritten by partial with the same name in the global layouts.
+-->
+<!-- Head custom content area end -->
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -52,4 +52,3 @@
 {{- template "partials/templates/twitter_cards.html" . }}
 {{- template "partials/templates/schema_json.html" . }}
 {{- end }}
-{{- partial "head_custom_content.html" -}}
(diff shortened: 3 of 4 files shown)


adityatelange:
> Hi! I was struggling with injecting custom code into on every page to enable analog of Google Analytics on the blog. The only way to do it in Hugo I found is to overwrite head.html or baseof.html partial from outside of the theme. However, it would require copying the code of the partial and updating it appropriately as the theme updates.
> 
> Another solution is to add to the theme a placeholder partial without any useful content into the baseof.html which in turn can be overwritten from outside of the theme. The same approach is already used in case of comments.html.
> 
> Please let me know if there is an easier way )

https://github.com/adityatelange/hugo-PaperMod/blob/7db1be3d7349eaba2dd87bfeb240b8bc120f4663/layouts/partials/head.html#L49-L54

We already have implemented google analytics code in it 
you just need to set `params.env` to production and add use this https://gohugo.io/templates/internal/#configure-google-analytics

Northerner-coder:
Thank you. However I need to use a different analytics service as well, which requires injecting custom code. This is why I had to find a different solution.

adityatelange:
> Thank you. However I need to use a different analytics service as well, which requires injecting custom code. This is why I had to find a different solution.

A hugo theme can be easily overwritten by custom implementation https://zwbetz.com/override-a-hugo-theme/, the user of the theme is creating extension to the theme by adding some code. Why not just clone `head.html` and make mods to it. People don't seem to update themes on their deployment on daily basis.

Northerner-coder:
The main drawback of this option is that it requires keeping the clone of head.html (or baseof.html, depending on the injection site) updated manually for it to be in sync with the original theme.  For instance, just yesterday you committed changes to head.html )

It looks like a valid mechanism, since it is used for injecting custom comments code without overwriting theme files per se. 

Of course, I will follow the route you mentioned if you believe that this change is not right for the theme and it's users.

layouts/_default/baseof.html
```diff
@@ -3,6 +3,7 @@
 
 <head>
     {{- partial "head.html" . }}
+    {{- partial "head_custom_content.html" -}}
```
adityatelange, line 6: Can you put this in head.html on L#48

adityatelange: and rename it to `extend_head.html`

adityatelange:
> The main drawback of this option is that it requires keeping the clone of head.html (or baseof.html, depending on the injection site) updated manually for it to be in sync with the original theme. For instance, just yesterday you committed changes to head.html )
I agree

layouts/_default/baseof.html
Northerner-coder: > Can you put this in head.html on L#48

and

> and rename it to `extend_head.html`

Done.  Plus added context . to partial call and changed slightly message in the extend_head.html placeholder file since it's content will not be added to the very end of <head> section any more.

singhkays:
This makes sense. I've seen similar approach in another theme. The partial there is called extra.html

PR #48, opened 2020-10-18
Localization for months

adityatelange:
* this fetches months translations for a language
* .Site.Params.months can be different for each language

PR #50, opened 2020-10-19
Months translation for archives.html via i18n

Northerner-coder:
Implemented translation of months in archives.html via i18n.  Months' translations can be used in other places, if necessary, as they are just pairs January="Translated name".  If chosen language does not contain translation of the chosen month it defaults to what Hugo uses as array key in archives.html - so nothing should change for current users.

Added translations for English and Russian.  Russian translation has also cleaned up a little bit: double quotes for consistency and counting options for translations (Hugo 0.76.5, maybe earlier started warning if such options are not present).

### DIFF
--- a/i18n/en.yaml
+++ b/i18n/en.yaml
@@ -14,3 +14,39 @@
 
 - id: translations
   translation: "Translations"
+
+- id: January
+  translation: "January"
+
+- id: February
+  translation: "February"
+
+- id: March
+  translation: "March"
+
+- id: April
+  translation: "April"
+
+- id: May
+  translation: "May"
+
+- id: June
+  translation: "June"
+
+- id: July
+  translation: "July"
+
+- id: August
+  translation: "August"
+
+- id: September
+  translation: "September"
+
+- id: October
+  translation: "October"
+
+- id: November
+  translation: "November"
+
+- id: December
+  translation: "December"
--- a/i18n/ru.yaml
+++ b/i18n/ru.yaml
@@ -52,3 +52,6 @@
 
 - id: December
   translation: "Декабрь"
+
+- id: Tags
+  translation: "Теги"
--- a/i18n/ru.yaml
+++ b/i18n/ru.yaml
@@ -1,14 +1,54 @@
 - id: prev_page
-  translation: 'Предыдущая'
+  translation: "Предыдущая"
 
 - id: next_page
-  translation: 'Следующая'
+  translation: "Следующая"
 
 - id: read_time
-  translation: '{{ .Count }} мин'
+  translation: 
+    one: "{{ .Count }} мин"
+    other: "{{ .Count }} мин"
+    few: "{{ .Count }} мин"
+    many: "{{ .Count }} мин"
 
 - id: toc
-  translation: 'Оглавление'
+  translation: "Оглавление"
 
 - id: translations
-  translation: 'Переводы'
+  translation: "Переводы"
+
+- id: January
+  translation: "Январь"
+
+- id: February
+  translation: "Февраль"
+
+- id: March
+  translation: "Март"
+
+- id: April
+  translation: "Апрель"
+
+- id: May
+  translation: "Май"
+
+- id: June
+  translation: "Июнь"
+
+- id: July
+  translation: "Июль"
+
+- id: August
+  translation: "Август"
+
+- id: September
+  translation: "Сентябрь"
+
+- id: October
+  translation: "Октябрь"
+
+- id: November
+  translation: "Ноябрь"
+
+- id: December
+  translation: "Декабрь"
--- a/layouts/_default/archives.html
+++ b/layouts/_default/archives.html
@@ -13,7 +13,9 @@
   </h2>
   {{- range .Pages.GroupByDate "January" }}
   <div class="archive-month">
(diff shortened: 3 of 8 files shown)


Northerner-coder:
@adityatelange Hi! Any thoughts on this issue?

adityatelange:
> @adityatelange Hi! Any thoughts on this issue?

Sorry but I couldn't think of an better impl than #48.

I'm not with i18n method as it is not feasible.
Also #48 is not a good impl, but can be merged temporarily. 

Let me know if you have any other way :)

Northerner-coder:
> > @adityatelange Hi! Any thoughts on this issue?
> 
> Sorry but I couldn't think of an better impl than #48.
> 
> I'm not with i18n method as it is not feasible.
> Also #48 is not a good impl, but can be merged temporarily.
> 
> Let me know if you have any other way :)

Do not see any meaningful improvement for https://github.com/adityatelange/hugo-PaperMod/pull/48 , except, maybe using default function.  Although, I am not sure it will work properly if the array with months is not set at all.

I can give it a try, however, first I would like to understand the issue with i18n.  I do not have much experience with Hugo and building template, but I had relatively good track record with i18n based on my other projects.  It is not ideal by far and can be cumbersome, but any localization is a messy business.  Could you explain for my understanding why https://github.com/adityatelange/hugo-PaperMod/pull/48 approach is better/cleaner than i18n, please?

adityatelange:
> > Sorry but I couldn't think of an better impl than #48.
> > I'm not with i18n method as it is not feasible.
> > Also #48 is not a good impl, but can be merged temporarily.
> > Let me know if you have any other way :)
> 
> Do not see any meaningful improvement for #48 , except, maybe using `default` function. Although, I am not sure it will work properly if the array with months is not set at all.
> 
> I can give it a try, however, first I would like to understand the issue with i18n. I do not have much experience with Hugo and building template, but I had relatively good track record with i18n based on my other projects. It is not ideal by far and can be cumbersome, but any localization is a messy business. Could you explain for my understanding why #48 approach is better/cleaner than i18n, please?

Doesn't https://github.com/adityatelange/hugo-PaperMod/pull/48#commitcomment-43349971 look better ? and easier ? and is sorted ? than 
https://github.com/adityatelange/hugo-PaperMod/pull/50/files#diff-d57e09739aebffaedf6e8a1d875a3ffcdc4f168361b3c292f7f7b1665372e4c0R17-R52

---

Suppose I want a multilingual "Archive" in a Dummy Lang.. What would I do is just fill in the 1-12 month translation for dummy lang, rather than waiting for someone to push it into `i18n` with that jan, feb, .. mess of strings.
Isn't that faster ? and simple ?

Northerner-coder:
> Doesn't [#48 (comment)](https://github.com/adityatelange/hugo-PaperMod/pull/48#commitcomment-43349971) look better ? and easier ? and is sorted ? than
> https://github.com/adityatelange/hugo-PaperMod/pull/50/files#diff-d57e09739aebffaedf6e8a1d875a3ffcdc4f168361b3c292f7f7b1665372e4c0R17-R52

Config-wise it does look better.  However, IMHO it has some drawbacks:

1. User has to know about and manually set this localization in the config file.  In i18n it works out of the box by simply changing the language (provided the language files are updated).

2. It splits localization into two files and mixes it with more or less pure configuration nature of config.

3. It moves "inelegance" from i18n localization file to the template itself (the code is more complicated than simply calling i18n, although it can be remedied by using `default` function if it works in this context).


> Suppose I want a multilingual "Archive" in a Dummy Lang.. What would I do is just fill in the 1-12 month translation for dummy lang, rather than waiting for someone to push it into `i18n` with that jan, feb, .. mess of strings.
> Isn't that faster ? and simple ?

Before I pushed extended version of the Russian translation to this repo, I created an i18n folder and file for Russian language in the global context which has superseded theme's localization file based on full English translation.  It took me 5 min and did not require any specific knowledge of the theme.  This can be done for any dummy language.  Moreover, it is a likely scenario, as an author will likely need to translate more than just an archives page.

adityatelange:
@Northerner-coder what are other themes using for this ? Has anyone else implemented it ?

Northerner-coder:
> @Northerner-coder what are other themes using for this ? Has anyone else implemented it ? like some themes have more than 1k users, this theme hasn't got that far yet.

I managed to find only one theme that dealt with translation of months: https://github.com/kakawait/hugo-tranquilpeak-theme .  Interestingly enough, they used combination of an [i18n file](https://github.com/kakawait/hugo-tranquilpeak-theme/blob/907866f6dbe76970371750cae96a4462cb0d21a4/i18n/en-us.yaml) and [scripted template](https://github.com/kakawait/hugo-tranquilpeak-theme/blob/907866f6dbe76970371750cae96a4462cb0d21a4/layouts/partials/internal/date.html).

Northerner-coder:
@adityatelange I guess you do not believe that this change is needed in the theme?  If not, I suggest closing this pr.

The latest commit is mistake: I use it in my fork to extend localization, committed to a wrong branch.  If this pr is closed, do not think that it should be a problem.

adityatelange:
> @adityatelange I guess you do not believe that this change is needed in the theme? If not, I suggest closing this pr.
Yes. It's half baked thus not gonna merge as of now 
 
---

> The latest commit is mistake: I use it in my fork to extend localization, committed to a wrong branch. If this pr is closed, do not think that it should be a problem.

okay

PR #51, opened 2020-10-19
chore: add go.mod

zhangyuannie:
This will allow https://gohugo.io/hugo-modules/use-modules/#use-a-module-for-a-theme without the `incompatible` tag.

So a user can simply write:
```toml
# config.toml
theme = "github.com/theNewDynamic/gohugo-theme-ananke"
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,3 @@
-module adityatelange/hugo-PaperMod
+module github.com/adityatelange/hugo-PaperMod
 
 go 1.12
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,3 @@
+module adityatelange/hugo-PaperMod
+
+go 1.12


adityatelange:
> This will allow https://gohugo.io/hugo-modules/use-modules/#use-a-module-for-a-theme without the `incompatible` tag.
> 
> So a user can simply write:
> 
> ```toml
> # config.toml
> theme = "github.com/theNewDynamic/gohugo-theme-ananke"
> ```

Nice add-on :+1:

go.mod
```diff
@@ -0,0 +1,3 @@
+module adityatelange/hugo-PaperMod
```
adityatelange, line 1: Shouldn't it be 
`module github.com/adityatelange/hugo-PaperMod` ?

zhangyuannie: Yes!

PR #54, opened 2020-10-22
Language switcher menu

adityatelange:
- Adds a List of Available Language other than current used one


adityatelange:
Preview:

image

image

PR #47, opened 2020-10-18
Add RTL support for multilingual mode

mehdy:
* Replace all instances of
  * `margin-left` to `margin-inline-start`
  * `margin-right` to `margin-inline-end`
  * `padding-left` to `padding-inline-start`
  * `padding-right` to `padding-inline-end`
* Remove `.pagination .next` and added `justify-content: space-between;` to `.pagination` instead.
* Add `direction-ltr` and `direction-rtl` classes.
* Assign one of the classes to `html` based on the [multilingual configuration](https://gohugo.io/content-management/multilingual/#configure-languages)

PR #57, opened 2020-10-23
Replace left with start for better RTL support

mehdy:
These are related to #47.

### DIFF
--- a/assets/css/post-single.css
+++ b/assets/css/post-single.css
@@ -132,7 +132,7 @@
 
 .post-content table th {
     font-size: 14px;
-    text-align: left;
+    text-align: start;
 }
 
 .post-content table:not(.highlighttable) td code:only-child {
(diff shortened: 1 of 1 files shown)


adityatelange:
@mehdy This patch is Fine..
Is there anything else left for RTL support ? Can you please check once with your config ?

mehdy:
@adityatelange Sure. I've done it already. works fine on both RTL and LTR modes.
Well, to be honest, I search for `left` and `right` in all of the styles and these were the only ones left to change.
So I think this all we need for now.

adityatelange:
> @adityatelange Sure. I've done it already. works fine on both RTL and LTR modes.
> Well, to be honest, I search for `left` and `right` in all of the styles and these were the only ones left to change.
> So I think this all we need for now.

Great !

PR #58, opened 2020-10-24
Feature/new translations

drvy:
Added the read_time, toc and translations for Spanish and create new translations for Bulgarian.


adityatelange:
@drvy Thank You for improving translations :)

PR #68, opened 2020-11-03
Broken description due to extra quote

ITAYC0HEN:
Hugo fails to build the website with the config in the Installation post as it has an extra quote in the "description".
I think that "ExampleSite description" is a nice compromise


adityatelange:
That is just an example config, that should not be used as it is.

Thanks again tho, I missed it :p

ITAYC0HEN:
Haha yeah, I figured. But a user should at least be able to build it as-is instead of troubleshooting it.

PR #69, opened 2020-11-03
Enable theme toggle unless specifically disabled

ITAYC0HEN:
Users can specify one of these three theme modes: Auto, Dark, Light.
In addition, there is a configuration option named `disableThemeToggle` that when disabled (i.e Theme toggle is enabled) allows the user to specify what theme they prefer to use, regardless of their system theme.

Currently, even when this config is disabeld (means that the Theme Toggle should be visible), the toggle would not be visible unless the author of the blog specified they want t use the "Auto" theme. That means, that if I, as a website owner, want the default theme to be `Light` and also allow the visitors to toggle the theme, I can't do this.

This Pull Request fixes this and honor `disableThemeToggle` regardles of the selected theme in the configuration. This mean, that the website owner have more control over whether the toggle will be shown or not.

### DIFF
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -63,7 +63,7 @@
     })
 </script>
 {{- else -}}
-<!-- case where owner disables theme button after deployment, this resets the stored theme -->
+{{/*  case where owner disables theme button after deployment, this resets the stored theme  */}}
 <script>
     localStorage.removeItem("pref-theme");
 </script>
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -50,7 +50,7 @@
         localStorage.setItem("menu-scroll-position", document.getElementById('menu').scrollLeft);
     }
 </script>
-{{- if (and (not .Site.Params.disableThemeToggle) (not (or (eq .Site.Params.defaultTheme "light") (eq .Site.Params.defaultTheme "dark")))) }}
+{{- if (not .Site.Params.disableThemeToggle) }}
 <script>
     document.getElementById("theme-toggle").addEventListener("click", () => {
         if (document.body.className.includes("dark")) {
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -1,15 +1,11 @@
-{{- if (and (not .Site.Params.disableThemeToggle) (not (or (eq .Site.Params.defaultTheme "light") (eq .Site.Params.defaultTheme "dark")))) }}
+{{- if (not .Site.Params.disableThemeToggle) }}
 <script>
     // load memory
     if (localStorage.getItem("pref-theme") === "dark") {
         document.body.classList.add('dark');
     } else if (localStorage.getItem("pref-theme") === "light") {
         document.body.classList.remove('dark')
-    } else {
-        if (window.matchMedia('(prefers-color-scheme: dark)').matches) {
-            document.body.classList.add('dark');
-        }
-    }
+    } 
 
 </script>
 {{- end }}
@@ -41,7 +37,7 @@
                 {{- .Site.Params.label.text | default .Site.Title -}}
             </a>
             <span class="logo-switches">
-                {{- if (and (not .Site.Params.disableThemeToggle) (not (or (eq .Site.Params.defaultTheme "light") (eq .Site.Params.defaultTheme "dark")))) }}
+                {{- if (not .Site.Params.disableThemeToggle) }}
                 <span class="theme-toggle">
                     <a id="theme-toggle" accesskey="t">
                         <svg id="moon" xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24"
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -1,17 +1,32 @@
-{{- if (not .Site.Params.disableThemeToggle) }}
+{{- if (eq .Site.Params.defaultTheme "light") }}
 <script>
-    // load memory
     if (localStorage.getItem("pref-theme") === "dark") {
         document.body.classList.add('dark');
     } else if (localStorage.getItem("pref-theme") === "light") {
         document.body.classList.remove('dark')
-    } 
+    } else {
+        document.body.classList.remove('dark')
+    }
 
 </script>
-{{- end }}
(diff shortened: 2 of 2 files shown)


adityatelange:
> Users can specify one of these three theme modes: Auto, Dark, Light.
> In addition, there is a configuration option named `disableThemeToggle` that when disabled (i.e Theme toggle is enabled) allows the user to specify what theme they prefer to use, regardless of their system theme.

Correct

---

> Currently, even when this config is disabeld (means that the Theme Toggle should be visible),

by config is disabled I take it as both vars `defaultTheme` and `disableThemeToggle` aren't set  or are commented out.

> the toggle would not be visible unless the author of the blog specified they want t use the "Auto" theme. 

That is not true. For both vars `defaultTheme` and `disableThemeToggle` not defined , the toggle is visible !
and will set to dark/light a/c to system theme. And if user clicks on the theme button it will store as preferred theme as light/dark.

> That means, that if I, as a website owner, want the default theme to be `Light` and also allow the visitors to toggle the theme, I can't do this.
The default theme will always be light unless system theme is dark. 

And yes if you set a default theme to **light / dark** it will disable theme toggle as site owner has preferred it that way.

---
related :

6a9e9e129a46f1b6ebdc7622dcf0519b7a582d4e

a9e00a3640b5d98e6fce03dc5c67e87e3fac916d

ITAYC0HEN:
Hence this PR brings additional flexibility. Now, I as a website owner can specify default them **AND** allow the visitor to choose their preferred one.

As I metioned in the example above. I want the users to present with a Light theme by default, regardless their browser theme, and at the same time, I want them to have the power to change theme by themselves to dark, if they want. 


The PR will give more flexibility to the owner, as can be seen by the following table:

| defaultTheme | disableThemeToggle | Preview |
|--------------|--------------------|---------|
| auto  (browser dark)       | true            |       image  |
| auto     (browser dark)      | false           |   image |
| light        | true            |        image |
| light        | false           |  image |
| dark         | true            | image  |
| dark         | false           |  image    |

adityatelange:
So this means 
You want to keep dark theme as **default** one instead of **auto** 
and let the user change accordingly ?

adityatelange:
This does break auto theme a/c to system theme !

Also imo if you are allowing theme to be changed i.e. user preferred, why not just let user's system theme decide the dark/ light ?
system theme is the user's preferred  theme anyways. 
So I think it already works are it should !

ITAYC0HEN:
For example :)
You can see a site I recently made with Hugo and LoveIt theme: https://explained.re
When you visit it, you will see a Dark theme, regardless of your browser theme. But as a visitor, you can still toggle it.

Why?
- Because I prefer my visitors to see dark theme (in explained.re)
- Some visitors would prefer the Light theme

so I want my website to be with one theme **and** provide them the option to change it. Because why not :)


>  system theme is the user's preferred theme anyways.

It is not necessarily. I use Arch Linux with Dark theme, and Direfox with Light theme and Twitter with Dark theme (and VSCode with Light theme...).

adityatelange:
@ITAYC0HEN Loveit theme is wayyy better than papermod !

I'll see if I can have a better solution to this with system theme to work as will as default theme :) will let you know ...

ITAYC0HEN:
In my solution, it works flawlessly, at least from what I tried.
I have a dark theme on my machine, and I want all my visitors to get the Light them first. So what I do is to set the default to "Light" and keep the Theme Toggle visible. So now everyone will see my Light theme, and still be able to change the theme to Dark if they wish to :)

adityatelange:
@ITAYC0HEN ping

adityatelange:
Sorry I don't know what happened, was working fine when I pushed then build didn't work as it should :/
Weird stuff

ITAYC0HEN:
Indeed doesn't work for me :)
Let me know when to test again.

adityatelange:
@ITAYC0HEN Can you please check b8157a2c79568ff674294d50ad0b27754bab74b6

ITAYC0HEN:
It doesn't work for me. Just to make sure we're on the same page -- I am with a dark-themed browser and with the following config:

```
  defaultTheme: light
  disableThemeToggle : false
```


I do get a light theme when visiting the website, but I don't have the toggle button
image

adityatelange:
> It doesn't work for me. Just to make sure we're on the same page -- I am with a dark-themed browser and with the following config:
> 
> ```
>   defaultTheme: light
>   disableThemeToggle : false
> ```
> 
> 
> I do get a light theme when visiting the website, but I don't have the toggle button
> image
> 

Hmm.. that's a problem, I think I need rest :/

Can you try using emulated dark/light theme in dev-tools, that is how I have tested (in chrome)

ITAYC0HEN:
From my tests, the original PR I sent works fine. Can you share a test plan in which it does not?

adityatelange:
@ITAYC0HEN 
This is how it should function basically

| `defaultTheme` | `disableThemeToggle` | check local storage? | check system theme? | Info              |
| -------------- | -------------------- | -------------------- | ------------------- | ----------------- |
| auto           | true                 | No                   | Yes                 | only system theme |
|                | false                | Yes (if not 1->2)    | Yes (2)             | switch present    |
| dark           | true                 | No                   | No                  | force dark only   |
|                | false                | Yes                  | No                  | switch present    |
| light          | true                 | No                   | No                  | force light only  |
|                | false                | Yes                  | No                  | switch present    |

adityatelange:
Please check if https://github.com/adityatelange/hugo-PaperMod/tree/theme-toggle-impr works fine. 
According to my checks it should work properly :)

ITAYC0HEN:
Fantastic! Looks great! :+1: 

This table can go to the docs as well :)
Good work!

adityatelange:
> Fantastic! Looks great! 
> 
> This table can go to the docs as well :)
> Good work!

@ITAYC0HEN So does every config work properly now ?

ITAYC0HEN:
Yeah I tested and it works as expected :)

adityatelange:
> Yeah I tested and it works as expected :)

So should I push it to this pr ?

ITAYC0HEN:
You can force push the commits to this PR so we will author it together

ITAYC0HEN:
Good work :)
Are you on Telegram or similar IM app? Have some questions&suggestions

adityatelange:
> Good work :)
> Are you on Telegram or similar IM app? Have some questions&suggestions

Sure, I'll contact you :) on tg

PR #79, opened 2020-11-18
social-icons: add svg for keybase

Kartoffel:
From @dabura667 https://github.com/keybase/keybase-issues/issues/973#issuecomment-307547386

@adityatelange related comment: would you be interested in adding a `lower` somewhere in the social_icons or svg partial? It would help in getting the capitalization for each service right (e.g. YouTube) while still getting the correct svg.


adityatelange (commented):
@Kartoffel keybase Icon is cool but I see it is 35 kb or so ! which is too large for an svg.. can we compress it ?
Also references in code is not good IMO, maybe add that in commit message ?
Also if dabura667 is author, why not give authorship ?

Kartoffel:
@adityatelange 
> @Kartoffel keybase Icon is cool but I see it is 35 kb or so ! which is too large for an svg.. can we compress it ?

It's purely points on the path, which is quite complex. Compressing here would be reducing paths, I've tried doing that automated but it left the logo distorted. I'm not a designer, so re-drawing the paths is a bit out of the question.

> Also references in code is not good IMO, maybe add that in commit message ?

It's in the commit message as well, you can remove it from the code if you want.

> Also if dabura667 is author, why not give authorship ?

Not sure how I could do that, do you have any info?

adityatelange (changes requested):
@Kartoffel 

can you add authorship(https://makandracards.com/makandra/1717-git-change-author-of-a-commit)
rm the comment reference and 
position it alphabetically ?

Kartoffel:
I managed to reduce the SVG to 2.5kB, and added dabura667 as co-author.
The reference has been removed from the code, and it's placed alphabetically now.

I see that in the mean time you've added lowercasing to the icon names, thanks!

adityatelange (approved):

adityatelange:
@Kartoffel Thank you :))

PR #82, opened 2020-11-22
social-icons: add svg for Mastodon

gikari:
from: https://joinmastodon.org/press-kit.zip

close adityatelange#81

Previews:

image
image


adityatelange:
@gikari Why not use https://joinmastodon.org/static/media/logo_transparent_white.6a441138.svg
simpler path right ?

Also can you arrange it alphabetically according to its name ?

gikari:
> Why not use https://joinmastodon.org/static/media/logo_transparent_white.6a441138.svg
> simpler path right ?
I thought it would not fit, since it's not wire-styled, but since you do not mind, I will change it to that.

adityatelange:
> > Why not use https://joinmastodon.org/static/media/logo_transparent_white.6a441138.svg
> > simpler path right ?
> 
> I thought it would not fit, since it's not wire-styled, but since you do not mind, I will change it to that.

Nope its Fine ! :+1:

adityatelange:
@gikari Thank you for contributing :)

gikari:
> @gikari Thank you for contributing :)

You're welcome :)

adityatelange:
@gikari It would be great if you could add icon to https://github.com/adityatelange/hugo-PaperMod/blob/exampleSite/content/posts/papermod/papermod-icons.md

PR #83, opened 2020-11-22
social-icons: add mastodon

gikari:
While is references mastodon.social, it can be any mastodon instance, e.g. mastodon.online, fosstodon.org etc


adityatelange:
@gikari Thank you :))

Issue #80, opened 2020-11-22
how to add-on mathjax?


pacientes:
I read some articles for hugo & mathjax..
But, I can't do that.

Could you help me?

pacientes:
I found [Math Typesetting](https://adityatelange.github.io/hugo-PaperMod/posts/math-typesetting/) post, and resolved it.

pacientes:
hello,

I configure my blog global config file like this.
![image](https://user-images.githubusercontent.com/22834091/100068654-27e36080-2e7b-11eb-9237-4a9c2bc4d974.png)

but math script was not working
![image](https://user-images.githubusercontent.com/22834091/100068818-519c8780-2e7b-11eb-840f-c1c99dd726c3.png)

the only i get success method is..
1. create math.html file in layouts/partials/ directory
2. add enable math script in extend_head.html
3. add math: true tag in specific posts markdown file (not global config)

Am I using the above method correctly?

adityatelange:
@pacientes 

Add `math: true` to post param in `somepost.md` not site `config.yml`
and add `enableInlineShortcodes: true` to site `config.yml`

PR #85, opened 2020-11-24
Add KaTeX html file

pacientes:
I want to include the KaTeX html file by default


adityatelange:
@pacientes Have you seen https://adityatelange.github.io/hugo-PaperMod/posts/math-typesetting/

and https://github.com/adityatelange/hugo-PaperMod/blob/exampleSite/config.yml#L5

pacientes:
oh... noooo...
sorry for not taking a closer look...

adityatelange:
> oh... noooo...
> sorry for not taking a closer look...

there was already a closed issue on this #80

PR #86, opened 2020-11-24
Improve theme switch logic

adityatelange:
- removes unnecessary local-storage item removal
- less inline script on build
- assumes `defaultTheme` as `auto` when value is not `dark` or `light`

should work according to table at [docs](https://adityatelange.github.io/hugo-PaperMod/posts/papermod/papermod-features/#theme-switch-toggle-enabled-by-default)

### DIFF
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -65,10 +65,4 @@
     })
 
 </script>
-{{- else -}}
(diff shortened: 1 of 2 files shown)


adityatelange:
@ITAYC0HEN Can you please check the updated logic ..

ITAYC0HEN (approved):
Works as expected. Good work :)

PR #90, opened 2020-11-27
Use internal templates where possible

philip-khor:
Use internal templates (https://gohugo.io/templates/internal/) for meta tags in the header where available.


adityatelange:
@philip-khor Those templates are modified specially for supporting cover images and other custom meta for papermod.
Meaning, it is intentional and not a mistake ! see commit history for more info

philip-khor:
My bad, sorry!

PR #105, opened 2020-12-03
Add Search

adityatelange:
- basic idea : https://gist.github.com/cmod/5410eae147e4318164258742dd053993
- uses fuse.js v6.4.3
- searches posts for only current Language
- `search.md`'s `url` should be 1 level up `/` i.e. directly under `content` dir

### DIFF
--- a/assets/js/fastsearch.js
+++ b/assets/js/fastsearch.js
@@ -41,7 +41,10 @@ function fetchJSONFile(path, callback) {
 // load our search index, only executed once
 function loadSearch() {
     fetchJSONFile('../index.json', function (data) {
-        var options = { // fuse.js options; check fuse.js website for details
+        // fuse.js options; check fuse.js website for details
+        var options = {
+            isCaseSensitive: true,
+            useExtendedSearch: true,
             shouldSort: true,
             location: 0,
             distance: 100,
--- a/assets/js/fastsearch.js
+++ b/assets/js/fastsearch.js
@@ -47,8 +47,7 @@ function fetchJSONFile(path, callback) {
 
 // load our search index, only executed once
 function loadSearch() {
-    fetchJSONFile('/index.json', function (data) {
-
+    fetchJSONFile('../index.json', function (data) {
         var options = { // fuse.js options; check fuse.js website for details
             shouldSort: true,
             location: 0,
--- a/assets/js/fastsearch.js
+++ b/assets/js/fastsearch.js
@@ -1,61 +1,57 @@
 var fuse; // holds our search engine
 
+// load our search index, only executed onload
+function loadSearch() {
+    var xhr = new XMLHttpRequest();
+    xhr.onreadystatechange = function () {
+        if (xhr.readyState === 4) {
+            if (xhr.status === 200) {
+                var data = JSON.parse(xhr.responseText);
+                if (data) {
+                    // fuse.js options; check fuse.js website for details
+                    var options = {
+                        isCaseSensitive: false,
+                        shouldSort: true,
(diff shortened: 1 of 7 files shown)


kdkasad:
How do you get the `index.json` file to generate? I've added the following to my `config.yaml` but no luck:
```
outputs:
  home:
    - html
    - rss
    - json
```

assets/js/fastsearch.js
```diff
@@ -0,0 +1,67 @@
...
+            }
+        }
+    };
+    httpRequest.open('GET', path);
+    httpRequest.send();
+}
+
+// load our search index, only executed once
+function loadSearch() {
+    fetchJSONFile('/index.json', function (data) {
```
kdkasad, line 50: This will fail if the site is in a subdirectory.

Example:  
If my site is `https://example.org/blog/`, the correct path for the JSON is `https://example.org/blog/index.json`, but the script will try to fetch `https://example.org/index.json`, which fails.

kdkasad: Fixed by #107

adityatelange:
@kdkasad It's WIP ! 
I am aware that wouldn't work .
I know you are eager to contribute, have patience buddy :))

kdkasad:
> @kdkasad It's WIP !
> I know you are eager to contribute, have patience buddy :))

Yep, I know. I just figured why not try to fix it since this is a feature I'd quite like, so I did. After all, it's one less problem for you to have to fix.

assets/js/fastsearch.js
adityatelange: is fixed by eb63b31

adityatelange:
## Sample Look and Feel
image

adityatelange:
> How do you get the `index.json` file to generate? I've added the following to my `config.yaml` but no luck:
> ```
> outputs:
>   home:
>     - html
>     - rss
>     - json
> ```
> 
> Edit: turns out the JSON is fine, but the search script is looking at the wrong URL.

Can you verify whether everything is well ?
It's ready to merge ig !

kdkasad:
> > How do you get the `index.json` file to generate? I've added the following to my `config.yaml` but no luck:
> > ```
> > outputs:
> >   home:
> >     - html
> >     - rss
> >     - json
> > ```
> > 
> > 
> > Edit: turns out the JSON is fine, but the search script is looking at the wrong URL.
> 
> Can you verify whether everything is well ?
> It's ready to merge ig !

Yes, it appears to be fine. The only problem is it would fail if the search page is not a direct subfolder of the `baseURL`. E.g. if baseURL is `https://example.org/my-site/` and the search page is `https://example.org/my-site/sub1/search/`, it will try to fetch `https://example.org/my-site/sub1/index.json`, which doesn't exist.

I don't think this would be a serious issue, and if it is, then it's easy to fix (see #107 which uses `{{ "index.json" | relURL}}` to get the URL for `index.json`.

adityatelange:
@kdkasad I am assuming here that search should be directly under content dir.
Otherwise it is not possible to support multilingual config !

Do you have any workaround for this ?

kdkasad:
> Otherwise it is not possible to support multilingual config !
> 
> Do you have any workaround for this ?

I think `{{ "index.json" | relLangURL }}` would generate the correct URL. I'll do some testing then get back to you.

kdkasad:
> I think `{{ "index.json" | relLangURL }}` would generate the correct URL. I'll do some testing then get back to you.

It doesn't, since `search.js` is only generated once, rather than once per language. I think your method is the best solution, so everything looks good to me.

PR #106, opened 2020-12-03
Make README non-executable

kdkasad:
There's no need for README.md to be executable


adityatelange:
@kdkasad What's the motive here ?

kdkasad:
> @kdkasad What's the motive here ?

Readme isn't an executable, so the executable permission should be disabled. All attempts to execute the file will fail anyways. It doesn't really matter (no security holes or anything), it's just a good practice.

PR #107, opened 2020-12-03
Obey baseURL when fetching index.json

kdkasad:
This fixes the problem I mentioned [here](https://github.com/adityatelange/hugo-PaperMod/pull/105#discussion_r535585084), where `fastsearch.js` always tries to fetch `/index.json`


kdkasad:
Isn't going to work for multiple languages without some hacky workarounds.

PR #122, opened 2020-12-10
Fix appending of '/' to menu item URLs

kdkasad:
Currently, menu item URLs automatically have a slash (`/`) appended to them. This is problematic if the resource they point to isn't supposed to have a trailing slash.

For example, on my website, I want a link to the RSS feed, which is located at `/blog/index.xml`. I set that as the URL for the menu item, but when generating, a slash (`/`) was appended, which turned it into `/blog/index.xml/`. This path is incorrect, and leads to a 404.

With this patch, menu item URLs don't have a slash appended. I tested it on same-site urls (like `/post/` and `/index.xml`) as well as full URLs (like `https://example.org`). Nothing seems to break.

### DIFF
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -109,7 +109,7 @@
         {{- $currentPage := . }}
         <ul class="menu" id="menu" onscroll="menu_on_scroll()">
             {{- range .Site.Menus.main }}
-            {{- $menu_item_url := (cond (strings.HasSuffix .URL "/") .URL (printf "%s/" .URL) ) | absLangURL }}
+            {{- $menu_item_url := .URL | absLangURL }}
             {{- $page_url:= $currentPage.Permalink  | absLangURL }}
             <li>
                 <a href="{{ $menu_item_url }}">


adityatelange:
>  Nothing seems to break.

Menu indicator does

kdkasad:
> > For example, on my website, I want a link to the RSS feed, which is located at `/blog/index.xml`. I set that as the URL for the menu item, but when generating, a slash (`/`) was appended, which turned it into `/blog/index.xml/`. This path is incorrect, and leads to a 404.
> 
> Haven't experienced this, try it on hugo's server command

I did. It has the same problem. Here's the relevant section of my `config.yaml`:
```
menu:
  main:
  - identifier: tags
    name: tags
    url: /tags/
    weight: 20
  - identifier: categories
    name: categories
    url: /categories/
    weight: 10
  - identifier: rss
    name: rss
    url: /index.xml
    weight: 30
  - identifier: search
    name: search
    url: /search/
    weight: 40
```

The `rss` menu entry will point to `/index.xml/`, when it should be `/index.xml`.

singhkays:
> For example, on my website, I want a link to the RSS feed, which is located at /blog/index.xml

🙏 Thank you! This is the exact problem I had and had to remove the slash using `sed` in by build script!

PR #113, opened 2020-12-07
fix custom taxonomy urls

sunuazizrahayu:
fix issue #112 
- allow custom permalink for taxonomies (tested on categories and tags)

example:
```
domain.com/tags/ #show all tags
domain.com/tags/foo -> domain.com/tag/foo  #for specific tag
```

PR #125, opened 2020-12-11
Fix URLs of tags in tag lists

kdkasad:
Solves the issue originally discussed in #113 using the solution [posted here](https://github.com/adityatelange/hugo-PaperMod/pull/113#issuecomment-742979288).

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -45,11 +45,10 @@
     {{- partial "anchored_headings.html" .Content -}}
   </div>
   <footer class="post-footer">
-    {{- if .Params.tags }}
+    {{- with .Params.tags }}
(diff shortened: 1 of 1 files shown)


layouts/_default/single.html
```diff
@@ -45,11 +45,10 @@ <h1 class="post-title">
     {{- partial "anchored_headings.html" .Content -}}
   </div>
   <footer class="post-footer">
-    {{- if .Params.tags }}
+    {{- with .Params.tags }}
```
adityatelange, line 48: do we need `with` ?

kdkasad: Nope. Changed in 3eed39d

PR #127, opened 2020-12-11
Don't use h4 tag in figure captions

### LINKED ISSUE #124: Image captions are added to the table of contents
If an image is added using the figure shortcode like so, the caption gets added to the page's Table of Contents:
```html
{{< figure src="/img/photo1.jpg" title="This is the caption" >}}
```

The resulting entry in the page's table of contents looks like this:
```html
<a href="#" aria-label="This is the caption">This is the caption</a>
```

I'm pretty sure this happens because the caption is put in an `h4` element:
```html
<figure>
    <img src="/very-secret-english-website/homemade-skis-broken.webp"/>
    <figcaption>
        <h4>and after</h4>
    </figcaption>
</figure>
```

kdkasad:
Fixes #124 by avoiding `h4` tags in `figcaption` elements. I've added to `post-content.css` so the look of captions should not change.

### DIFF
--- a/assets/css/post-single.css
+++ b/assets/css/post-single.css
@@ -224,6 +224,13 @@
     border-radius: 4px
 }
 
+.post-content figure > figcaption {
+    color: var(--primary);
+    font-size: 16px;
+    font-weight: bold;
+    margin: 24px 0 16px;
+}
+
 .toc {
     margin-bottom: 40px;
     border: 1px solid var(--border);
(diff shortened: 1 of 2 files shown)


adityatelange:
@kdkasad Now that we are including it in theme itself, how about customizing it and make it look like the cover image ?

kdkasad:
> how about customizing it and make it look like the cover image ?

How exactly? The embedded images already look like the cover image.

PR #132, opened 2020-12-14
[WIP] Include KaTeX assets in theme

kdkasad:
### Changes
Includes assets (scripts, stylesheets, and fonts) required for KaTeX math markup.

This would replace the method described in [math-typesetting.md on the example site](https://adityatelange.github.io/hugo-PaperMod/posts/math-typesetting/), so that should either be updated with new instructions or removed if this gets merged.

### Usage
KaTeX can be enabled site-wide by adding the following to `config.yaml`:
```yaml
params:
  katex: true
```
It can also be enabled for a specific page by including `katex: true` in the page's front matter (this is the recommended way).

Once it's been enabled, you can include KaTeX markup in your content.
Example:
```markdown
Inline math: \(\varphi = \dfrac{1+\sqrt5}{2}= 1.6180339887…\)

Block math:

$$
 \varphi = 1+\frac{1} {1+\frac{1} {1+\frac{1} {1+\cdots} } }
$$
```

### Problems
This includes lots of fonts in the repository (in `static/fonts/`). @adityatelange Let me know how you feel about this. I don't know of a way around this ATM.


adityatelange:
@kdkasad This is as simple as https://github.com/bee-san/hugo-PaperMod/commit/c5c9ffc3006f5fb499b2a8c66b94bdd3800433eb

Although this needs to be done in site and not necessarily in theme itslef, 
`extend_head.html` is best possible way to do this

kdkasad:
> @kdkasad This is as simple as [bee-san@c5c9ffc](https://github.com/bee-san/hugo-PaperMod/commit/c5c9ffc3006f5fb499b2a8c66b94bdd3800433eb)
> 
> Although this needs to be done in site and not necessarily in theme itslef,
> `extend_head.html` is best possible way to do this
> 
> I had to include fuze.js inline because it can be widely used by anybody, and not just Maths enthusiast.
> I do not intend to not accredit your efforts on this, but it is always a good practice to create an issue/discussion and ask to be assigned on it. Hope you understand :)

Got it, all of that makes sense.

Issue #130, opened 2020-12-14
Specifying canonical URL results in multiple canonical URLs


cbrownstein:
_Originally posted by @adityatelange in https://github.com/adityatelange/hugo-PaperMod/issues/104#issuecomment-738547278_:
> @Horsty80 It's fine to have multiple canonical urls. Canonical urls just let search engines know that their contents are same. You may read about those on the web, for more info :)

I believe this is incorrect and can cause problems.

See, e.g.:

https://yoast.com/rel-canonical/#common-issues
> Multiple rel=canonical links on a page cause havoc. When we encounter this in WordPress plugins, we try to reach out to the developer doing it and teach them not to, but it still happens. And when it does, the results are wholly unpredictable.

https://help.dragonmetrics.com/en/articles/214017-multiple-rel-canonical-tags
> If there is more than one rel=canonical tag on the page, search engines will get confused and not know which tag contains the actual canonical page. If there is more than one on the page, Google will simply ignore all of the rel=canonical tags, while other search engines may behave erratically or unpredictably when trying to identify the canonical page.

Horsty80:
@cbrownstein i'm totaly agree with you, and i custom the theme to have just 1 canonical URL
Thanks for the ressources ;)

kdkasad:
I don't think this is an actual issue. The theme doesn't create multiple canonical URLs. The problem was only caused by adding more custom content to the `head` element.

Horsty80:
My workaround is to overload the `head` element

cbrownstein:
> I don't think this is an actual issue. The theme doesn't create multiple canonical URLs. The problem was only caused by adding more custom content to the `head` element.

The theme _can_ create multiple canonical URLs.

The relevant lines are:

https://github.com/adityatelange/hugo-PaperMod/blob/09aba49bf0922110bd61b23a802b2099ce7b5a36/layouts/partials/head.html#L19-L23

Line 19 will output the canonical URL of the post _always_.

Lines 20-23 will output one (or more) user-defined canonical URLs if `.Site.Params.canonical` has values or `.Params.canonical` has values.

kdkasad:
> > I don't think this is an actual issue. The theme doesn't create multiple canonical URLs. The problem was only caused by adding more custom content to the `head` element.
> 
> The theme _can_ create multiple canonical URLs.
> 
> The relevant lines are:
> 
[...]

PR #104, opened 2020-12-03
update Canonical For External Url

Horsty80:
Add custom front matter variable called canonicalUrl
That allow the user add external canonical url.
For example user that write article for his job or on other blogin platform but he want to add the article on his own website

PR #133, opened 2020-12-15
Fix user defining canonical URL resulting in multiple canonical URLs

cbrownstein:
With 8af7c55: If the user defines a canonical URL, `head` will have multiple canonical URLs, which is improper. So, I'm reverting that commit in 6649bef.

With 12ea911: If the user defines a canonical URL, that URL will be used instead of the URL of the post.

N.B. This PR is basically the same as #104 which was opened by @Horsty80.

Closes: #130


kdkasad:
I don't think this would work. Looking at the original code, it seems like `.Params.canonical` is an array (or whatever Hugo calls it) that must be iterated over using `range`.

adityatelange:
@cbrownstein I'll just add the check for home page, if you don't mind..

PR #136, opened 2020-12-15
Include menu information in example site

kdkasad:
Adds a menu section to the sample `config.yml`, as well as a section on menu entries in the How-To Guide post.

In the sample menu, I added two in-site links and one external URL to show that both are possible.


adityatelange:
@kdkasad you are fast !

PR #140, opened 2020-12-16
Add mathjax support

ataias:
Ports with the `hasMath` variable set to true on the front-matter have
mathjax added to the page.

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -71,6 +71,7 @@
 <link rel="alternate" hreflang="{{- .Lang -}}" href="{{- .Permalink -}}" />
 {{ end }}
 {{- partial "extend_head.html" . -}}
+{{ partial "mathjax" . }}
 <!-- Misc -->
 {{- if hugo.IsProduction | or (eq .Site.Params.env "production") }}
 {{- template "_internal/google_analytics_async.html" . }}
--- a/layouts/partials/mathjax.html
+++ b/layouts/partials/mathjax.html
@@ -1,4 +1,4 @@
-{{ if .Params.hasMath }}
+{{ if or .Params.mathjax .Site.Params.mathjax }}
 <script
   type="text/javascript"
   async
--- a/layouts/partials/mathjax.html
+++ b/layouts/partials/mathjax.html
@@ -1,0 +1,32 @@
+{{ if .Params.hasMath }}
+<script
+  type="text/javascript"
+  async
+  src="https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML"
+>
+  MathJax.Hub.Config({
+  tex2jax: {
+    inlineMath: [['$','$'], ['\\(','\\)']],
+    displayMath: [['$$','$$']],
+    processEscapes: true,
+    processEnvironments: true,
+    skipTags: ['script', 'noscript', 'style', 'textarea', 'pre'],
+    TeX: { equationNumbers: { autoNumber: "AMS" },
+         extensions: ["AMSmath.js", "AMSsymbols.js"] }
+  }
+  });
+  MathJax.Hub.Queue(function() {
+    // Fix <code> tags after MathJax finishes running. This is a
+    // hack to overcome a shortcoming of Markdown. Discussion at
+    // https://github.com/mojombo/jekyll/issues/199
+    var all = MathJax.Hub.getAllJax(), i;
+    for(i = 0; i < all.length; i += 1) {
+        all[i].SourceElement().parentNode.className += ' has-jax';
+    }
+  });
+  MathJax.Hub.Config({
+  // Autonumbering by mathjax
+  TeX: { equationNumbers: { autoNumber: "AMS" } }
+  });
+</script>
+{{ end }}


adityatelange:
@kdkasad what do you feel about this one as compared to #132

adityatelange:
@ataias Mathjax better or katex ? coz I see https://github.com/bee-san/hugo-PaperMod/commit/c5c9ffc3006f5fb499b2a8c66b94bdd3800433eb this as easier implementation

kdkasad:
> @kdkasad what do you feel about this one as compared to #132

I agree with your decision to not include MathJax or KaTeX as part of the theme. In this PR, it has to fetch external sources, which isn't great. In #132, the KaTeX assets are included in the theme, which adds a lot of files.

If I had to choose one though, I'd do a combo of both and include a KaTeX partial which fetches the external assets (basically bee-san@c5c9ffc). KaTeX seems a bit simpler to implement IMO.

adityatelange:
> I agree with your decision to not include MathJax or KaTeX as part of the theme. In this PR, it has to fetch external sources, which isn't great. In #132, the KaTeX assets are included in the theme, which adds a lot of files. Again, not great.
> 
> If I had to choose one though, I'd do a combo of both and include a KaTeX partial which fetches the external assets (basically [bee-san/hugo-PaperMod@c5c9ffc](https://github.com/bee-san/hugo-PaperMod/commit/c5c9ffc)). KaTeX seems a bit simpler to implement IMO.

Exactly !

adityatelange:
@kdkasad maybe we should add https://github.com/bee-san/hugo-PaperMod/commit/c5c9ffc this to 
https://adityatelange.github.io/hugo-PaperMod/posts/math-typesetting/

ataias:
> @ataias Mathjax better or katex ? coz I see [bee-san@c5c9ffc](https://github.com/bee-san/hugo-PaperMod/commit/c5c9ffc3006f5fb499b2a8c66b94bdd3800433eb) this as easier implementation

This PR enables one to use math with latex symbol for indicating the equations. For instance, [this page](https://ataias.com.br/2016/09/03/code-a-program-to-solve-a-partial-differential-equation-using-finite-differences-with-an-explicit-method/) whose source code in markdown is like:

```markdown
Before coding anything, we need to expand the `$\nabla f(x,y)$` term using
finite differences on it. To expand the term is quite easy and results in
`\begin{align} \nabla f(x, y) &= \frac{\partial^2 f}{\partial x^2} + \frac{\partial^2 f}{\partial y^2}. \end{align}`

Finite differences is a method that allows us to remove the continuous `$x$` and
`$y$` variables to introduce discrete variables `$i$` and `$j$` which will be
used to iterate in a loop. This is an approximation and the accuracy increases
with the number of discrete points one wants to use, but makes the
implementation more complicated as well as costs more computing power. To
discretize a second order partial differential term with second order accuracy,
we have the relation
`\begin{align} \frac{\partial^2 f}{\partial x^2}(x, y) &\approx \frac{f(x_{i-1},y_i)-2f(x_{i}, y_i)+f(x_{i+1},y)}{\Delta x^2}, \end{align}`
where `$\Delta x = \mathrm{span}(x)/N$` and `$N$` is the number of points that a
given dimension has. Remember that the continuous variable `$x$` has infinite
points, but when we use `$x_i$` we use only a finite set of points of size
`$N$`. The number of points `$N$` will be same for both dimensions, so we have
`$\Delta y = \Delta x$`.

Regarding `$u(x,y)$` we just evaluate it at `$(i\Delta x, j\Delta x)$`. This
way, we have our discrete equation

<!-- prettier-ignore -->
`\begin{align}
c_{ij} = \sum_{m=-1,1}f(x_{i-m}, y_i) + \sum_{n=-1,1}f(x_i, y_{j-n}),\\
\frac{1}{\Delta x^2}\left(c_{ij} - 4f(x_i, y_i)\right) = u(i\Delta x, j\Delta x).\label{eq:discretizedPoisson}
\end{align}`

Notice that Equation `\eqref{eq:discretizedPoisson}` is our discretized Poisson
equation. As the focus of this post is to present code using an explicit method,
we isolate the term `$f_{ij}$` like in
`\begin{align} f_{ij} &= \frac{c_{ij}}{4} -\frac{u_{ij}\Delta x^2}{4}. \label{eq:internalPoints} \end{align}`

We can use Equation `\eqref{eq:internalPoints}` for any point internal to our
`$1\times 1$` square. Nevertheless, for the boundary points, i.e where
`$x\in \{0,1\}$` or `$y\in\{0,1\}$`, we can not use it as we will not be able to
compute `$i-1$`, `$i+1$`, `$j-1$` or `$j+1$`. I mean, these points would be
outside our mesh.
```

With the katex implementation in https://github.com/bee-san/hugo-PaperMod/commit/c5c9ffc , the formulas are not rendered. They are not rendered with just a line including mathjax either. 

I am not sure about the feature difference between katex and mathjax, but I imagine there is a difference, so I think supporting both would be nicer.

bee-san:
> katex and mathjax, but I imagine there is a difference, so I think supporting both



> > @ataias Mathjax better or katex ? coz I see [bee-san@c5c9ffc](https://github.com/bee-san/hugo-PaperMod/commit/c5c9ffc3006f5fb499b2a8c66b94bdd3800433eb) this as easier implementation
> 
> This PR enables one to use math with latex symbol for indicating the equations. For instance, [this page](https://ataias.com.br/2016/09/03/code-a-program-to-solve-a-partial-differential-equation-using-finite-differences-with-an-explicit-method/) whose source code in markdown is like:
> 
> ```gfm
> Before coding anything, we need to expand the `$\nabla f(x,y)$` term using
> finite differences on it. To expand the term is quite easy and results in
> `\begin{align} \nabla f(x, y) &= \frac{\partial^2 f}{\partial x^2} + \frac{\partial^2 f}{\partial y^2}. \end{align}`
> 
> Finite differences is a method that allows us to remove the continuous `$x$` and
> `$y$` variables to introduce discrete variables `$i$` and `$j$` which will be
> used to iterate in a loop. This is an approximation and the accuracy increases
> with the number of discrete points one wants to use, but makes the
> implementation more complicated as well as costs more computing power. To
> discretize a second order partial differential term with second order accuracy,
> we have the relation
> `\begin{align} \frac{\partial^2 f}{\partial x^2}(x, y) &\approx \frac{f(x_{i-1},y_i)-2f(x_{i}, y_i)+f(x_{i+1},y)}{\Delta x^2}, \end{align}`
> where `$\Delta x = \mathrm{span}(x)/N$` and `$N$` is the number of points that a
> given dimension has. Remember that the continuous variable `$x$` has infinite
> points, but when we use `$x_i$` we use only a finite set of points of size
> `$N$`. The number of points `$N$` will be same for both dimensions, so we have
> `$\Delta y = \Delta x$`.
> 
> Regarding `$u(x,y)$` we just evaluate it at `$(i\Delta x, j\Delta x)$`. This
> way, we have our discrete equation
> 
> 
> `\begin{align}
> c_{ij} = \sum_{m=-1,1}f(x_{i-m}, y_i) + \sum_{n=-1,1}f(x_i, y_{j-n}),\\
> \frac{1}{\Delta x^2}\left(c_{ij} - 4f(x_i, y_i)\right) = u(i\Delta x, j\Delta x).\label{eq:discretizedPoisson}
> \end{align}`
> 
> Notice that Equation `\eqref{eq:discretizedPoisson}` is our discretized Poisson
> equation. As the focus of this post is to present code using an explicit method,
> we isolate the term `$f_{ij}$` like in
> `\begin{align} f_{ij} &= \frac{c_{ij}}{4} -\frac{u_{ij}\Delta x^2}{4}. \label{eq:internalPoints} \end{align}`
> 
> We can use Equation `\eqref{eq:internalPoints}` for any point internal to our
> `$1\times 1$` square. Nevertheless, for the boundary points, i.e where
> `$x\in \{0,1\}$` or `$y\in\{0,1\}$`, we can not use it as we will not be able to
> compute `$i-1$`, `$i+1$`, `$j-1$` or `$j+1$`. I mean, these points would be
> outside our mesh.
> ```
> 
> With the katex implementation in [bee-san@c5c9ffc](https://github.com/bee-san/hugo-PaperMod/commit/c5c9ffc) , the formulas are not rendered. They are not rendered with just a line including mathjax either.
> 
> I am not sure about the feature difference between katex and mathjax, but I imagine there is a difference, so I think supporting both would be nicer.

Hey!

Generally, I don't use equations in codeblocks, but maths renders fine otherwise :)

Having used both MathJax and Katex (at least for the last week, Katex) I think I prefer Katex. It renders a lot faster. Although my biggest gripe with MathJax was how much it broke, and how I had to write code to scan my blog to find the broken mathjax so I can fix it.

Although I agree that giving options is best :-)

adityatelange:
In my opinion I do not feel any of both (katex and mathjax) should be a part of theme.
As of now **no external resources are fetched** and it is a good approach IMO.

User already has _choice_ what extra needs to be added to personalize their own website, 
we also have `extend_head` and extend_footer` so as to add stuff like this !

Thank You
~ Aditya

ataias:
> In my opinion I do not feel any of both (katex and mathjax) should be a part of theme.
> As of now **no external resources are fetched** and it is a good approach IMO.
> 
> User already has _choice_ what extra needs to be added to personalize their own website,
> we also have `extend_head` and extend_footer` so as to add stuff like this !
> 
> Thank You
> ~ Aditya

In the case of `extend_head`, is it necessary to change the file in the theme folder or can it be outside?

ataias:
> > katex and mathjax, but I imagine there is a difference, so I think supporting both
> 
> > > @ataias Mathjax better or katex ? coz I see [bee-san@c5c9ffc](https://github.com/bee-san/hugo-PaperMod/commit/c5c9ffc3006f5fb499b2a8c66b94bdd3800433eb) this as easier implementation
> > 
> > 
> > This PR enables one to use math with latex symbol for indicating the equations. For instance, [this page](https://ataias.com.br/2016/09/03/code-a-program-to-solve-a-partial-differential-equation-using-finite-differences-with-an-explicit-method/) whose source code in markdown is like:
> > ```
> > Before coding anything, we need to expand the `$\nabla f(x,y)$` term using
> > finite differences on it. To expand the term is quite easy and results in
> > `\begin{align} \nabla f(x, y) &= \frac{\partial^2 f}{\partial x^2} + \frac{\partial^2 f}{\partial y^2}. \end{align}`
> > 
> > Finite differences is a method that allows us to remove the continuous `$x$` and
> > `$y$` variables to introduce discrete variables `$i$` and `$j$` which will be
> > used to iterate in a loop. This is an approximation and the accuracy increases
> > with the number of discrete points one wants to use, but makes the
> > implementation more complicated as well as costs more computing power. To
> > discretize a second order partial differential term with second order accuracy,
> > we have the relation
> > `\begin{align} \frac{\partial^2 f}{\partial x^2}(x, y) &\approx \frac{f(x_{i-1},y_i)-2f(x_{i}, y_i)+f(x_{i+1},y)}{\Delta x^2}, \end{align}`
> > where `$\Delta x = \mathrm{span}(x)/N$` and `$N$` is the number of points that a
> > given dimension has. Remember that the continuous variable `$x$` has infinite
> > points, but when we use `$x_i$` we use only a finite set of points of size
> > `$N$`. The number of points `$N$` will be same for both dimensions, so we have
> > `$\Delta y = \Delta x$`.
> > 
> > Regarding `$u(x,y)$` we just evaluate it at `$(i\Delta x, j\Delta x)$`. This
> > way, we have our discrete equation
> > 
> > 
> > `\begin{align}
> > c_{ij} = \sum_{m=-1,1}f(x_{i-m}, y_i) + \sum_{n=-1,1}f(x_i, y_{j-n}),\\
> > \frac{1}{\Delta x^2}\left(c_{ij} - 4f(x_i, y_i)\right) = u(i\Delta x, j\Delta x).\label{eq:discretizedPoisson}
> > \end{align}`
> > 
> > Notice that Equation `\eqref{eq:discretizedPoisson}` is our discretized Poisson
> > equation. As the focus of this post is to present code using an explicit method,
> > we isolate the term `$f_{ij}$` like in
> > `\begin{align} f_{ij} &= \frac{c_{ij}}{4} -\frac{u_{ij}\Delta x^2}{4}. \label{eq:internalPoints} \end{align}`
> > 
> > We can use Equation `\eqref{eq:internalPoints}` for any point internal to our
> > `$1\times 1$` square. Nevertheless, for the boundary points, i.e where
> > `$x\in \{0,1\}$` or `$y\in\{0,1\}$`, we can not use it as we will not be able to
> > compute `$i-1$`, `$i+1$`, `$j-1$` or `$j+1$`. I mean, these points would be
> > outside our mesh.
> > ```
> > 
> > 
> > With the katex implementation in [bee-san@c5c9ffc](https://github.com/bee-san/hugo-PaperMod/commit/c5c9ffc) , the formulas are not rendered. They are not rendered with just a line including mathjax either.
> > I am not sure about the feature difference between katex and mathjax, but I imagine there is a difference, so I think supporting both would be nicer.
> 
> Hey!
> 
> Generally, I don't use equations in codeblocks, but maths renders fine otherwise :)
> 
> Having used both MathJax and Katex (at least for the last week, Katex) I think I prefer Katex. It renders a lot faster. Although my biggest gripe with MathJax was how much it broke, and how I had to write code to scan my blog to find the broken mathjax so I can fix it.
> 
> Although I agree that giving options is best :-)
> 
> I will add that I am not good at frontend stuff at all, and the only time I touch frontend is to get something to work in an essay I've written so I'm definitely not the expert here. Just a writer!

I don't recall why I used those code blocks. I removed them locally and it worked too.

ataias:
> > In my opinion I do not feel any of both (katex and mathjax) should be a part of theme.
> > As of now **no external resources are fetched** and it is a good approach IMO.
> > User already has _choice_ what extra needs to be added to personalize their own website,
> > we also have `extend_head` and extend_footer` so as to add stuff like this !
> > Thank You
> > ~ Aditya
> 
> In the case of `extend_head`, is it necessary to change the file in the theme folder or can it be outside?

I found out how. I like this option, wasn't aware of it. Closing this.

PR #144, opened 2020-12-22
Add markdown image center align using CSS tag

pacientes:
Hello.
This is the markdown tag that centers the image.

How to use:
`name`

- Add `#center` tag

Result:
image


pacientes:
oh pr is wrong

PR #165, opened 2021-01-01
Add post description

weicliu:
Usage example already exists in the markdown-syntax.md: https://github.com/adityatelange/hugo-PaperMod/blame/exampleSite/content/posts/markdown-syntax.md#L5

### DIFF
--- a/assets/css/post-single.css
+++ b/assets/css/post-single.css
@@ -8,6 +8,11 @@
     transform: translateX(-2px)
 }
 
+.post-description {
+    margin-top: 10px;
+    margin-bottom: 5px;
+}
+
 .post-meta {
     color: var(--secondary);
     font-size: 14px;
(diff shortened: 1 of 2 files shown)


adityatelange (approved):

mukhtharcm:
But where does the description is shown?

kdkasad:
> But where does the description is shown?

Between the post title and the cover image:
screenshot

mukhtharcm:
Oh!,
Is there any way to show it on the post list page,?
Instead of showing  the starting lines of the post, it'd been cool if we can show a description instead

kdkasad:
> Oh!,
> Is there any way to show it on the post list page,?
> Instead of showing the starting lines of the post, it'd been cool if we can show a description instead

Currently, I don't think so, but it wouldn't be hard to implement. Try giving it a shot. Take a look at the changes made in this PR and try to do the same in `layouts/_default/list.html`.

mukhtharcm:
Thanks. 
I'll try it in the morning. 
It's 1.30 at night here 😀😀

Issue #126, opened 2020-12-11
[BUG] wrong load stylesheet and JS on multihost (multilingual)


sunuazizrahayu:
my config:
```
languages:
  id:
    baseURL: "site.com/id/"
    languageName: "Indonesia"
  en:
    baseURL: "site.com/en/"
    languageName: "English"
```
on preview of `hugo server`, second language (english) load stylesheet from first languange (indonesia).
it produce
![Capture](https://user-images.githubusercontent.com/18210880/101905502-268b9500-3bea-11eb-8ec0-df4dab3de360.PNG)
i think this config must use their own baseUrl.
so stylesheet must loaded as `http://site.com/en/my.css`, not `http://site.com/id/my.css`

---

Noted points:

1. <strike>Always uses 1st Language's (by alphabetical order) assets </strike>

adityatelange:
Style-sheet should be common ! Why do they need  to be loaded different for different language ?

sunuazizrahayu:
thanks..

adityatelange:
@sunuazizrahayu I actually feel this is hugo's bug !
1. head is not using `partialCached` so cached page shouldn't be used of some other lang.
2. The asset generated is using `Permalink` which means it is a permanent link to the asset.

sunuazizrahayu:
how about make this theme to use their own assets on every lang?

adityatelange:
> how about make this theme to use their own assets on every lang?

Wait, we'll research about this more ..

sunuazizrahayu:
when render, it's produce assets for every host separately
![Capture](https://user-images.githubusercontent.com/18210880/101910710-eb8d5f80-3bf1-11eb-9ae6-8799bd2673ab.PNG)

adityatelange:
@sunuazizrahayu 

Reading https://gohugo.io/content-management/multilingual/ and set up config like mentioned [here](https://gohugo.io/content-management/multilingual/#configure-multilingual-multihost)


![image](https://user-images.githubusercontent.com/21258296/102252233-d97e2a80-3f2b-11eb-86e0-4b5a4d7aeb9b.png)

![image](https://user-images.githubusercontent.com/21258296/102252260-e4d15600-3f2b-11eb-9d05-a73d723ea243.png)

---

![image](https://user-images.githubusercontent.com/21258296/102252924-b0aa6500-3f2c-11eb-91f1-5d83255eb41c.png)

[...]

PR #168, opened 2021-01-04
CORS issue for multidomain setup fixed adityatelange/hugo-PaperMod#126

wersoo:
Hi,

I stumbled upon an issue with the CORS similar to the #126 in multilingual mode with distinct domains assigned per language.
A similar case was described [here](https://github.com/gohugoio/hugo/issues/5226#issuecomment-438953402) and using `.RelPermalink` to load Hugo Pipes resources instead of just `Permalink` was the unanimously recommended workaround.

Tested it locally and on netlify, and it worked well.


adityatelange:
I always prefer absolute urls over relative urls, but this seems to be a nice workaround until the issue is fixed...
Also one can get abs urls by passing `canonifyURLs: true` in the config if they need it :)

Thanks !

Issue #172, opened 2021-01-11
ShowReadingTime in Bahasa Indonesia is broken


snydez:
Hi there,
When I set `defaultContentLanguage: id`  and `ShowReadingTime: true`  the display is broken.
While if I didn't set `defaultContentLanguage`, things are ok.

![image](https://user-images.githubusercontent.com/2133889/104143827-9f515d00-53f3-11eb-8be9-ce50d38bbbe7.png)

muhammadhanif:
Create `id.yaml` file and save it in `themes/hugo-PaperMod/i18n/`:

```
- id: prev_page
  translation: "Halaman Sebelumnya"

- id: next_page
  translation: "Halaman Selanjutnya"

- id: read_time
  translation:
    one : "1 menit"
    other: "{{ .Count }} menit"

- id: toc
  translation: "Daftar Isi"

- id: translations
  translation: "Terjemahan"
```

adityatelange:
> Hi there,
> When I set `defaultContentLanguage: id` and `ShowReadingTime: true` the display is broken.
> While if I didn't set `defaultContentLanguage`, things are ok.
> 
> ![image](https://user-images.githubusercontent.com/2133889/104143827-9f515d00-53f3-11eb-8be9-ce50d38bbbe7.png)

Sorry but I cannot reproduce the issue, checkout the snap below
![image](https://user-images.githubusercontent.com/21258296/104189111-3ba15100-5440-11eb-8d38-cb6e29582162.png)

with **id** in URL and `defaultContentLanguage` unset :thinking:

snydez:
Hi @adityatelange 
I try @muhammadhanif suggestion, and it's working now.


thank you all

PR #173, opened 2021-01-11
Add Translation: Bahasa Indonesia

muhammadhanif:
Hi, 

Based on #172 issue,  here I add a new translation for Bahasa Indonesia.

Thank you.


adityatelange:
@muhammadhanif Thanks for tranalations :+1: , 
but #172 should not display that kind of text even if there is no i18n file present.

adityatelange (approved):

PR #180, opened 2021-01-16
Prev next post links on single post page

adityatelange:
- In Site config: `Params.ShowPostNavLinks: true`

### DIFF
--- a/assets/css/post-single.css
+++ b/assets/css/post-single.css
@@ -382,7 +382,7 @@ h6:hover .anchor {
 .paginav {
     margin: 10px 0;
     display: flex;
-    line-height: 36px;
+    line-height: 30px;
 }
 
 .paginav .title {
--- a/assets/css/post-single.css
+++ b/assets/css/post-single.css
@@ -374,3 +374,38 @@ h6:hover .anchor {
 .post-content table::-webkit-scrollbar-thumb {
     border-width: 2px;
 }
+
+.post-content ::-webkit-scrollbar-thumb:hover {
+    background: rgba(255, 255, 255, 0.64);
+}
+
+.paginav {
+    margin: 10px 0;
+    display: flex;
+    line-height: 36px;
+}
+
+.paginav .title {
+    letter-spacing: 1px;
+    text-transform: uppercase;
+    font-size: small;
+    color: var(--secondary);
+}
+
+.paginav .prev,
+.paginav .next {
+    width: 50%;
+}
+
+.paginav span:hover:not(.title) {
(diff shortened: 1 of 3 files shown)


adityatelange:
## Demo

<kbd>image
</kbd>

<kbd>image</kbd>

kdkasad:
Looks great!

adityatelange:
> Looks great!

I still feel the position isn't great, and I am not having any other ideas to fix this.

kdkasad:
> > Looks great!
> 
> I still feel the position isn't great, and I am not having any other ideas to fix this.

I wonder if for the mobile layout, it might be better to omit the titles of the next/previous posts and instead use a button like the **Next Page** button from the homepage.

adityatelange:
> > > Looks great!
> > 
> > 
> > I still feel the position isn't great, and I am not having any other ideas to fix this.
> 
> I wonder if for the mobile layout, it might be better to omit the titles of the next/previous posts and instead use a button like the **Next Page** button from the homepage. That way they'd fit side-by-side.

Yes you are correct, I have tried it but it doesn't look good at all with long text titles and also with emoji's.
(On the Home Page the buttons aren't user editable, making it smaller)

muhammadhanif:
> ## Demo
> image
> 
> image

In my opinion, the prev-next post links would be nice if it is placed below of share buttons . Why? Because the share buttons is still part of the post (for sharing the post).

This is what I implement on my site.

screenshot-hanifmu com-2021 01 19-15_51_12

adityatelange:
> In my opinion, the prev-next post links would be nice if it is placed below of share buttons . Why? Because the share buttons is still part of the post (for sharing the post).
> 
> This is what I implement on my site.
> 
> screenshot-hanifmu com-2021 01 19-15_51_12
> 

The design does match #183 :)

muhammadhanif:
> > In my opinion, the prev-next post links would be nice if it is placed below of share buttons . Why? Because the share buttons is still part of the post (for sharing the post).
> > This is what I implement on my site.
> > screenshot-hanifmu com-2021 01 19-15_51_12
> 
> The design does match #183 :)

absolutely yes 😁

by the way, Thank you for releasing this great theme 👍

kdkasad:
> In my opinion, the prev-next post links would be nice if it is placed below of share buttons . Why? Because the share buttons is still part of the post (for sharing the post).
> 
> This is what I implement on my site.
> 
> screenshot-hanifmu com-2021 01 19-15_51_12

Now that I've seen both ways, I agree with this layout.

PR #192, opened 2021-01-23
Make "Powered by" translatable

fkohrt:


adityatelange:
Is this really needed ? IMO not.
Likewise we need to translate a ton of stuff, including dates and "Theme" string.

fkohrt:
What triggered this for me was that a localized `copyright` string clashed with the non-localized _Powered by_ in the footer. I thought about _Theme_, but that can be regarded as a fixed Hugo term. And localization of dates can already be achieved with

```yml
languages:
  en:
    languageName: English
    params:
      DateFormat: 'January 2, 2006'
```

Or what other localization did you have in mind?

adityatelange:
> What triggered this for me was that a localized `copyright` string clashed with the non-localized _Powered by_ in the footer and provides a bad experience with screen readers. I thought about _Theme_, but that could be regarded as a fixed Hugo term.

I agree

fkohrt:
@adityatelange Alright, I have accounted for that.

> Likewise we need to translate a ton of stuff

What probably still lacks translation are the accessible names such as `aria-label`. But that's for another PR...

RealA10N:
> Is this really needed ? IMO not.
> Likewise we need to translate a ton of stuff, including dates and "Theme" string.

I think that yes, it should indeed be translated. If someone is building a website in a language other than English, those little "patches" of English can be really annoying. In any case, strings *should not* be hardcoded.

PR #195, opened 2021-01-24
patch fingerprinting func in templates

hlwjia:
Other params instead of `md5` are `sha256`, `sha512`, `sha384` (from Hugo 0.55).

### DIFF
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -20,7 +20,7 @@
 {{- $isHLJSdisabled := (.Site.Params.assets.disableHLJS | default .Params.disableHLJS ) }}
 {{- if (and (eq .Kind "page") (ne .Layout "archives") (ne .Layout "search") (not $isHLJSdisabled)) }}
 {{- if not .Site.Params.assets.disableFingerprinting }}
-{{- $highlight := slice (resources.Get "js/highlight.min.js") | resources.Concat "assets/js/highlight.js" | minify | fingerprint }}
+{{- $highlight := slice (resources.Get "js/highlight.min.js") | resources.Concat "assets/js/highlight.js" | minify | resources.Fingerprint "md5" }}
 <script defer src="{{ $highlight.RelPermalink }}" integrity="{{ $highlight.Data.Integrity }}"
     onload="hljs.initHighlightingOnLoad();"></script>
 {{- else}}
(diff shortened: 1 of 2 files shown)


adityatelange:
@hlwjia The current usage is valid. 
Read https://gohugo.io/hugo-pipes/introduction/#method-aliases

Also the default one used is `sha256`, as you can read here: https://gohugo.io/hugo-pipes/fingerprint/

So this won't fix "failed to find a valid digest..." error

hlwjia:
@adityatelange I noticed that when I deployed a site of mine with your theme, I get the same hash from my css filename as yours at https://adityatelange.github.io/hugo-PaperMod/. 

Starts with `9fcc3f6...`

I thought maybe your `fingerprint` didn't work or something, so I did a quick search and found `resources.Fingerprint`. So I did a quick test with `md5`. It worked.

Now, after reading your reply. Thanks for that. I didn't know the method alias before. I got confused why it didn't work for me.

Then I tested with, again, `resources.Fingerprint  "sha256"`, which worked for me except using md5. 

Guess what! It doesn't work again.

And then I tested with `resources.Fingerprint "sha512"`. 

Surprisingly, it works!

Now I'm really curious why only `sha256` won't work for me while `md5` and `sha512` works perfectly.

hlwjia:
Just did another test with `fingerprint "sha512"`, works a like charm.

Any clue why only the sha256 didn't work?

adityatelange:
> Any clue why only the sha256 didn't work?


Maybe https://github.com/gohugoio/hugo/issues/7882 ? I'm not really sure but it sometimes does not recalculate, is what my experience is !

adityatelange (approved):
I think we should merge this since this also generates a smaller hash, which does the same job.

hlwjia:
Do you want me to change it back to use method alias `fingerprint` since it's not really the cause of the problem? 

If you want, I can update it to`fingerprint "md5"` on my branch. And then create a new pull request.

I actually prefer cleaner alias :D

adityatelange:
> Do you want me to change it back to use method alias `fingerprint` since it's not really the cause of the problem?
> 
> If you want, I can update it to`fingerprint "md5"` on my branch. And then create a new pull request.
> 
> I actually prefer cleaner alias :D

https://github.com/adityatelange/hugo-PaperMod/commit/9e99493e696961bf6dfd58c8540184c8ad4132f7#commitcomment-46358507

also this is an issue. 

I didn't read docs https://www.w3.org/TR/SRI/#cryptographic-hash-functions before merging.

It says `User agents should refuse to support known-weak hashing functions like MD5 or SHA-1`, idk why hugo supports it ..

hlwjia:
@adityatelange Thanks for the additional research. 

Sorry for causing any inconvenience. :(

PR #198, opened 2021-01-26
Specify input type of search field

fkohrt:
Recommendation as per https://ffoodd.github.io/a11y.css/errors.html#no-type
See also https://references.modernisation.gouv.fr/publications/rgaa-accessibilite/methode/criteres/#crit-11-11

### DIFF
--- a/layouts/_default/search.html
+++ b/layouts/_default/search.html
@@ -31,7 +31,7 @@
 </header>
 
 <div id="searchbox">
-    <input id="searchInput" autofocus placeholder="{{.Title}} ↵" aria-label="search">
+    <input type="search" id="searchInput" autofocus placeholder="{{.Title}} ↵" aria-label="search">
     <ul id="searchResults" aria-label="search results"></ul>
 </div>
 


adityatelange:
> Recommendation as per https://ffoodd.github.io/a11y.css/errors.html#no-type
> See also https://references.modernisation.gouv.fr/publications/rgaa-accessibilite/methode/criteres/#crit-11-11

1. I don't understand French.
2. It is same as text and search, as there is no Form being submitted rn.
3. This change adds a cross sign in the end for which a on click handler needs to be added.

fkohrt:
1. Nevermind, I provided the second (french) link only because those RGAA guidelines are cited in the first link but with a broken link.
2. I'm not sure about the importance of this commit accessibility wise, but it has been a suggestion of the [a11y.css](https://ffoodd.github.io/a11y.css/) tool so I thought I'd share.
3. I believe this is browser specific behaviour (Firefox doesn't show this), so not sure why and how a onclick handler should be added.

So feel free to use to close this if you don't consider it worth – not sure about this either.

fkohrt:
@adityatelange I don't see anything to improve this commit right now, or would you rather see the cross sign hidden?

adityatelange:
Sorry but I don't think we need this

PR #202, opened 2021-01-27
Add japanese traslations

Hattomo:
Thank you for your beautiful theme!

I added some Japanese translations.


adityatelange (approved):
Thanks

PR #222, opened 2021-02-07
make integrity checking optional

boehs:


boehs:
oops just realized this is already implimented by 

```yml
params:
    assets:
        disableFingerprinting: true
```

PR #247, opened 2021-02-17
Redesign image captions

### LINKED ISSUE #246: How to center the text in a caption?
Related to https://github.com/adityatelange/hugo-PaperMod/discussions/241. As suggested, I did use the provided shortcode `figure` & it "works" but not as intended.

I was trying to caption an image like this:
image

But the formatting of the text in the caption generated by Hugo is like this:
image

What am I doing wrong? Or am I missing some feature of your theme? I see you provide a customized version of the `figure` shortcode with `PaperMod`. So, perhaps I'm not using it the right way?

Any guidance will be much appreciated.

kdkasad:
Changes the style of image captions. This creates a distinciton between the `title` and `caption` attributes. The former now shows up more prominently than the latter. Both are centered below the image, which resolves #246.

Since there are no pages in the example site with images, I've added one in my fork's `exampleSite` branch: kdkasad@e7e62d98d4356c57bc9878e7cda8d5048f743f38. Below are screenshots of the various images on that page.

#### Edit: captions are no longer centered. See [this comment](https://github.com/adityatelange/hugo-PaperMod/pull/247#issuecomment-781897919) for updated screenshots.

Cover image with caption:
cover image

Content image with title, caption, and attribution:
image with title, caption, attr

Content image with title and caption:
image with title and caption

Content image with title:
image with title

Content image with caption:
image with caption

### DIFF
--- a/assets/css/post-single.css
+++ b/assets/css/post-single.css
@@ -246,7 +246,13 @@
     color: var(--primary);
     font-size: 16px;
     font-weight: bold;
-    margin: 24px 0 16px;
+    margin: 8px 0 16px;
+}
+
+.post-content figure > figcaption > p {
+	color: var(--secondary);
+	font-size: 14px;
+	font-weight: normal;
 }
 
 .toc {


adityatelange:
The issue I thought for customizing captions in #127 inside the post was Cover image is meant to be full sized and fits the width.

While in case of images inside a post, the images can be smaller than width of `main` hence centering the caption and title does not seem a nice idea to me.

Ex. 

image

kdkasad:
> The issue I thought for customizing captions in #127 inside the post was Cover image is meant to be full sized and fits the width.
> 
> While in case of images inside a post, the images can be smaller than width of `main` hence centering the caption and title does not seem a nice idea to me.
> 
> Ex.
> 
> image

I didn't think of that. Should I make the captions left-aligned as they were before, or should I try to center them to the image, even if the image is smaller than the page?

Jarmos-san:
Centering `figure` makes the most sense here. That way, even if the image is smaller in width than the rest of the page, it won't look weird for the reader.

adityatelange:
> Centering `figure` makes the most sense here. That way, even if the image is smaller in width than the rest of the page, it won't look weird for the reader.

I am sorry but this is your preference, not everyone wants this

kdkasad:
> > Centering `figure` makes the most sense here. That way, even if the image is smaller in width than the rest of the page, it won't look weird for the reader.
> 
> I am sorry but this is your preference, not everyone wants this

I completely understand. I've rebased the code and removed the center-alignment. Captions are now aligned according to the language's direction, which fixes the problems with images that are smaller than the page width. Here are some example images:

#### Left-to-right language direction
ltr

#### Right-to-left language direction
rtl

#### Right-to-left language direction with small image
rtl with small image

Jarmos-san:
Well, these image orientations works for me too! I wouldn't mind that. Hugo's default formatting for `figure` just looks bland for PaperMod is what I wanted to say. 

**Centering it wasn't specifically for my needs, but was rather a suggestion** I thought would look nice with the theme. Apologies if my suggestions sounded otherwise.

assets/css/post-entry.css
```diff
@@ -104,3 +104,7 @@
     color: var(--secondary);
     box-shadow: 0 1px 0 var(--primary)
 }
+
+.entry-cover p {
+	margin: 8px 0 16px;
+}
```
adityatelange, line 110: Cover seems fine to me. Can you please remove this

adityatelange (commented):
Everything else look good :)

assets/css/post-entry.css
kdkasad: I had added this to make the cover image captions have the same padding as  the other image captions. I'm fine with removing it, but are you sure we don't want this?

**With padding:**
caption with padding

**Without padding:**
caption without padding

adityatelange: yes

kdkasad: I removed the second commit. Should be good to go now.

adityatelange (approved):

PR #254, opened 2021-02-19
enforce language url subsitution on path boundary

oasys:
Without this, the section name gets mangled when the language was a
  substring of the section.  For example, under 'en', a section named
  "fragment" changes to "fragmt".

### DIFF
--- a/layouts/partials/breadcrumbs.html
+++ b/layouts/partials/breadcrumbs.html
@@ -1,7 +1,7 @@
 {{- if (.Param "ShowBreadCrumbs")}}
 <div class="breadcrumbs">
     {{- $url := replace .Parent.Permalink ( printf "%s" .Site.BaseURL) "" }}
-    {{- $lang_url := replace $url ( printf "%s" .Lang) "" }}
+    {{- $lang_url := strings.TrimPrefix ( printf "%s/" .Lang) $url }}
 
     <a href="{{ "" | absLangURL }}">{{ i18n "home" | default "Home"}}</a>
     {{- $scratch := newScratch }}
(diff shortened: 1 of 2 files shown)


Shaddee060314 (approved):
layouts/partials/templates/schema_json.html

adityatelange:
@oasys What if we use `[LIMIT]` from https://gohugo.io/functions/replace/ ? Does it have same effect

like

```diff
-{{- $lang_url := replace $url ( printf "%s/" .Lang) "/" }}
+{{- $lang_url := replace $url ( printf "%s/" .Lang) "/" 1 }}
```

oasys:
I will test later, but I suspect that will only fix when running a multi language config. A single language config will probably still have problems (as in my example, there will be no `en/` to strip. )

adityatelange:
> I will test later, but I suspect that will only fix when running a multi language config. A single language config will probably still have problems (as in my example, there will be no `en/` to strip. )

True

oasys:
How about this?   I assume multilingual config always has the lang prefix at the beginning of the URI.   Using `replaceRE` anchors the lang match to the beginning (`^`) and the end (`/`).     I think this should work for most cases -- the only one I can think of is having a "section" name that is the same as a configured language.

oasys:
This will do the same, but might be more readable:

```
 {{- $lang_url := strings.TrimPrefix ( printf "%s/" .Lang) $url }}
```

If you prefer this, let me know and I'll update the PR.

adityatelange:
> This will do the same, but might be more readable:
> 
> ```
>  {{- $lang_url := strings.TrimPrefix ( printf "%s/" .Lang) $url }}
> ```
> 
> If you prefer this change instead of the `replaceRE` one, let me know and I'll update the PR.

I agree, sure we can do this way

oasys:
> I agree, sure we can do this way

Done.  Thanks!

adityatelange (approved):

PR #256, opened 2021-02-20
Add 'align' option to figure shortcode

kdkasad:
Allows centering figures (images) by adding an `align` parameter to the `figure` shortcode.

Adding `align=center` to a figure causes the resulting figure to be centered.

Example screenshot:
image

### DIFF
--- a/assets/css/post-single.css
+++ b/assets/css/post-single.css
@@ -243,9 +243,6 @@
 }
 
 .post-content figure.align-center {
-	display: flex;
-	flex-direction: column;
-	align-items: center;
 	text-align: center;
 }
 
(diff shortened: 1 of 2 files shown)


kdkasad:
Actually now that I'm thinking about it, there's no need for an `align` parameter with a value of `center`, there could just be a `center` parameter which takes effect based on whether or not it's there.

E.g.
```go
{{< figure src=image.jpg center >}}
```
instead of
```go
{{< figure src=image.jpg align=center >}}
```

adityatelange:
@kdkasad I guess we can make it simpler with already implemented code

https://github.com/adityatelange/hugo-PaperMod/blob/45c4b48e53ab8b9054b6ed6da2c72b17ef4de690/assets/css/post-single.css#L241-L243

PR #269, opened 2021-02-27
Do not show ToC when it does not exist

pvtmert:
Hello, I've noticed that ToC shows up with an empy section when there are no headings exist and ShowToc is set to true.

I've fixed problem by adding `and` operator to `single.html` template. However I'm not sure what's the ideal option there. If something else needed, please point me out.

Best.

Before:
image

After:
image


adityatelange:
@pvtmert Why not use `showtoc = false` for that page in front matter vars

https://github.com/adityatelange/hugo-PaperMod/wiki/Variables#page-variables

pvtmert:
Thank you for pointing that out. Have a nice week!

PR #274, opened 2021-02-28
Add support for Font Awesome

zincnode:
# [Font Awesome](https://fontawesome.com/)

>Get vector icons and social logos on your website with Font Awesome, the web's most popular icon set and toolkit.

## Add global support for Font Awesome

Adding script to *head.html* makes Font Awesome globally available.

### [Usage](https://fontawesome.com/how-to-use/on-the-web/referencing-icons/basic-use)

Example:

  ```html
<i class="fas fa-camera"></i>
  ```

## Add short code support for Font Awesome

Create *fa.html* to add support for Font Awesome shortcodes in markdown.

>via: [Font Awesome in Hugo](https://matze.rocks/posts/fontawesome_in_hugo/)

### [Usage](https://matze.rocks/posts/fontawesome_in_hugo/#4-usage)

Example:

  ```html
{{< fa bath >}}
  ```

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -83,3 +83,5 @@
 {{- template "partials/templates/twitter_cards.html" . }}
 {{- template "partials/templates/schema_json.html" . }}
 {{- end }}
+<!-- Font Awesome -->
+<script defer src="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.15.2/js/all.min.js"></script>
--- a/layouts/shortcodes/fa.html
+++ b/layouts/shortcodes/fa.html
@@ -1,0 +1,14 @@
+<!-- use default fontawesome class 'fas' or use a specified one like 'fab' for brands -->
+{{ $faStylePrefix := "fas" }}
+{{ if in (slice "fas" "far" "fal" "fad" "fab") (.Get 0) }}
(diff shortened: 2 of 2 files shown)


adityatelange:
I am not sure if we need this. 
There are no external scripts/styles are loaded from anywhere in PaperMod and imho we'll stick to that..

This can be simply placed in [extended head/footer](https://adityatelange.github.io/hugo-PaperMod/posts/papermod/papermod-faq/#custom-head--footer).

zincnode:
> I am not sure if we need this.
> There are no external scripts/styles are loaded from anywhere in PaperMod and imho we'll stick to that..
> 
> This can be simply placed in [extended head/footer](https://adityatelange.github.io/hugo-PaperMod/posts/papermod/papermod-faq/#custom-head--footer).

It's up to you to decide whether to change it or not. 😃 I need some icons in font awesome, but I don't know if others need them.

Both [all.min.js](https://github.com/FortAwesome/Font-Awesome/blob/master/js/all.min.js) and [all.min.css](https://github.com/FortAwesome/Font-Awesome/blob/master/css/all.min.css) I use can be saved inside the theme.But I think loading from outside through CDN may make loading faster.

adityatelange:
Sorry but I would like to keep this at it is, it will affect theme's performance and don't want this!

PR #278, opened 2021-03-01
Add option to show link in meta data for editing posts

### LINKED ISSUE #238: [Feature] Edit button for posts
I was looking for an option to easily add the ability to suggest changes to posts. I saw another hugo theme did this [here](https://github.com/stapelberg/hugo/blob/947e73e83246507891f687c201ee7be0bb47adb9/themes/stapelberg/layouts/_default/single.html#L41) resulting in [this](https://michael.stapelberg.ch/posts/2012-09-23-systemd-ionice/).

I'd suggest making it optionally by specifying an `editUrl`:
```
params:
  editUrl: "https://github.com/stapelberg/hugo/edit/main/"
```

And adding a edit button/string to `layouts/_default/single.html`:
```
{{- if (.Param "editUrl") }}
  <a href="{{ .Param "editUrl" }}/content/{{ .File.Path }}">Suggest changes</a>
{{- end }}
```
I believe the originally used `.File.LogicalName` is not what is needed here since it does not account for subfolders.

I'M not good as designing stuff so it looks good or I would have opened a PR. If this is something that is wanted and someone could give me a hand with where to put it, I#d be happy to open an PR.

Syphdias:
The idea was inspired by the [stapelberg theme].

If `editPostUrl` is set, its value will be used to create a link for every post
is displayed in the meta data of the post (if meta not hidden). This is useful
for providing the reader with an easy way to find the proper place to suggest
changes. An example would be an edit prefix for the proper git repo, e.g.:
```
Params:
  editPostUrl: "https://github.com/adityatelange/hugo-PaperMod/edit/exampleSite"
```
For the post saved in `content/posts/markdown-syntax.md`, this will lead to an
href of
"https://github.com/adityatelange/hugo-PaperMod/edit/exampleSite/content/posts/markdown-syntax.md"

The text for the edit button defaults to "Edit" or a translation if available
but can be changed individually by setting the parameter `editPostText`.

This closes #238.

PS: There is one issue that is not being addressed by this PR and that is docs – and I don't think I can directly solve this in this PR.
Where is documentations supposed to go?
* `README.md` on branch  `master`?
* Wiki of this repo?
* `content/posts/papermod/papermod-variables.md` on branch `exampleSite`
* Multiple places? (<-wrong answer)

Also:
* Before changes are merged?
* After changes are merged?
* At the same time?

[stapelberg theme]: https://github.com/stapelberg/hugo/blob/d0b2a3edb61a855d87debc3d974ed250853aa931/themes/stapelberg/layouts/_default/single.html#L46

### DIFF
--- a/layouts/partials/edit_post.html
+++ b/layouts/partials/edit_post.html
@@ -1,0 +1,4 @@
+{{- if (.Param "editPost.URL") -}}
+{{- if or .Params.author $.Site.Params.author (.Param "ShowReadingTime") (not .Date.IsZero) }}|&nbsp;{{- end -}}
+<a href="{{ .Param "editPost.URL" }}/{{ .File.Path }}">{{ .Param "editPost.Text" | default (i18n "edit_post" | default "Edit") }}</a>
+{{- end }}
--- a/layouts/partials/edit_post.html
+++ b/layouts/partials/edit_post.html
@@ -1,3 +1,0 @@
-{{- if (.Param "editPostUrl") }}
-&nbsp;·&nbsp;<a href="{{ .Param "editPostUrl" }}/content/{{ .File.Path }}">{{ .Param "editPostText" | i18n "edit_post" | default "Edit" }}</a>
-{{- end }}
--- a/layouts/partials/edit_post.html
+++ b/layouts/partials/edit_post.html
@@ -1,0 +1,3 @@
+{{- if (.Param "editPostUrl") }}
+&nbsp;·&nbsp;<a href="{{ .Param "editPostUrl" }}/content/{{ .File.Path }}">{{ .Param "editPostText" | i18n "edit_post" | default "Edit" }}</a>
+{{- end }}
--- a/layouts/partials/edit_post.html
+++ b/layouts/partials/edit_post.html
@@ -1,4 +1,4 @@
-{{- if (.Param "editPost.URL") -}}
-{{- if or .Params.author $.Site.Params.author (.Param "ShowReadingTime") (not .Date.IsZero) }}|&nbsp;{{- end -}}
-<a href="{{ .Param "editPost.URL" }}/{{ .File.Path }}">{{ .Param "editPost.Text" | default (i18n "edit_post" | default "Edit") }}</a>
+{{- if or .Params.editPost.URL .Site.Params.editPost.URL -}}
+{{- if or .Params.author $.Site.Params.author (.Param "ShowReadingTime") (not .Date.IsZero) .IsTranslated }}|&nbsp;{{- end -}}
+<a href="{{ .Params.editPost.URL | default .Site.Params.editPost.URL }}/{{ .File.Path }}" rel="noopener noreferrer" target="_blank">{{ .Params.editPost.Text | default (.Site.Params.editPost.Text | default (i18n "edit_post" | default "Edit") ) }}</a>
 {{- end }}
--- a/layouts/partials/edit_post.html
+++ b/layouts/partials/edit_post.html
@@ -1,4 +1,6 @@
 {{- if or .Params.editPost.URL .Site.Params.editPost.URL -}}
 {{- if or .Params.author $.Site.Params.author (.Param "ShowReadingTime") (not .Date.IsZero) .IsTranslated }}|&nbsp;{{- end -}}
-<a href="{{ .Params.editPost.URL | default .Site.Params.editPost.URL }}/{{ .File.Path }}" rel="noopener noreferrer" target="_blank">{{ .Params.editPost.Text | default (.Site.Params.editPost.Text | default (i18n "edit_post" | default "Edit") ) }}</a>
+<a href="{{ .Params.editPost.URL | default .Site.Params.editPost.URL }}{{ if or .Params.editPost.appendFilePath .Site.Params.editPost.appendFilePath | default true }}/{{ .File.Path }}{{ end }}" rel="noopener noreferrer" target="_blank">
+    {{- .Params.editPost.Text | default (.Site.Params.editPost.Text | default (i18n "edit_post" | default "Edit") ) -}}
+</a>
 {{- end }}
--- a/layouts/partials/edit_post.html
+++ b/layouts/partials/edit_post.html
@@ -1,6 +1,6 @@
 {{- if or .Params.editPost.URL .Site.Params.editPost.URL -}}
 {{- if or .Params.author $.Site.Params.author (.Param "ShowReadingTime") (not .Date.IsZero) .IsTranslated }}|&nbsp;{{- end -}}
-<a href="{{ .Params.editPost.URL | default .Site.Params.editPost.URL }}{{ if or .Params.editPost.appendFilePath .Site.Params.editPost.appendFilePath | default true }}/{{ .File.Path }}{{ end }}" rel="noopener noreferrer" target="_blank">
+<a href="{{ .Params.editPost.URL | default .Site.Params.editPost.URL }}{{ if .Params.editPost.appendFilePath | default ( .Site.Params.editPost.appendFilePath | default false ) }}/{{ .File.Path }}{{ end }}" rel="noopener noreferrer" target="_blank">
     {{- .Params.editPost.Text | default (.Site.Params.editPost.Text | default (i18n "edit_post" | default "Edit") ) -}}
 </a>
 {{- end }}
--- a/i18n/de.yaml
+++ b/i18n/de.yaml
@@ -14,3 +14,6 @@
 
 - id: translations
   translation: "Übersetzungen"
+
+- id: edit_post
+  translation: "Bearbeiten"
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -16,7 +16,6 @@
     <div class="post-meta">
       {{- partial "post_meta.html" .  -}}
       {{- partial "translation_list.html" . -}}
-      {{- partial "edit_post.html" . -}}
     </div>
     {{- end}}
   </header>
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -16,6 +16,7 @@
     <div class="post-meta">
       {{- partial "post_meta.html" .  -}}
       {{- partial "translation_list.html" . -}}
+      {{- partial "edit_post.html" . -}}
     </div>
     {{- end}}
   </header>
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -15,8 +15,8 @@
     {{- if not (.Param "hideMeta") }}
     <div class="post-meta">
       {{- partial "post_meta.html" .  -}}
-      {{- partial "edit_post.html" . -}}
       {{- partial "translation_list.html" . -}}
+      {{- partial "edit_post.html" . -}}
     </div>
     {{- end}}
   </header>
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -15,6 +15,7 @@
     {{- if not (.Param "hideMeta") }}
     <div class="post-meta">
       {{- partial "post_meta.html" .  -}}
+      {{- partial "edit_post.html" . -}}
       {{- partial "translation_list.html" . -}}
     </div>
     {{- end}}
--- a/layouts/partials/post_meta.html
+++ b/layouts/partials/post_meta.html
@@ -12,7 +12,7 @@
 {{ $scratch.Add "meta" (slice .)}}
 {{end}}
 
-{{- if (.Param "editPost") }}
+{{- if (.Param "editPost.URL") }}
 {{ $scratch.Add "meta" (printf "<a href=\"%s/content/%s\">%s</a>" (.Param "editPost.URL") .File.Path (.Param "editPost.Text" | i18n "edit_post" | default "Edit") ) }}
 {{- end -}}
 
--- a/layouts/partials/post_meta.html
+++ b/layouts/partials/post_meta.html
@@ -12,6 +12,10 @@
 {{ $scratch.Add "meta" (slice .)}}
 {{end}}
 
+{{- if (.Param "editPost") }}
+{{ $scratch.Add "meta" (printf "<a href=\"%s/content/%s\">%s</a>" (.Param "editPost.URL") .File.Path (.Param "editPost.Text" | i18n "edit_post" | default "Edit") ) }}
+{{- end -}}
+
 {{- with ($scratch.Get "meta")}}
 {{- delimit . "&nbsp;·&nbsp;"}}
 {{- end }}
--- a/layouts/partials/post_meta.html
+++ b/layouts/partials/post_meta.html
@@ -13,7 +13,7 @@
 {{end}}
 
 {{- if (.Param "editPost.URL") }}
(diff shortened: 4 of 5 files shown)


layouts/partials/edit_post.html
```diff
@@ -0,0 +1,3 @@
+{{- if (.Param "editPostUrl") }}
+&nbsp;·&nbsp;<a href="{{ .Param "editPostUrl" }}/content/{{ .File.Path }}">{{ .Param "editPostText" | i18n "edit_post" | default "Edit" }}</a>
```
kdkasad, line 2: I think that `/content/{{ .File.Path}}` should not be appended to the value of `editPostUrl`. For example, if someone wanted to use a `mailto` link, it would not make sense to append a path to the email address.

adityatelange (changes requested):
1. How about adding this to `post_meta` partial so that it would not display extra `·` when other meta params are empty 
2. It would be appropriate to keep `editPost` params with each other
    like shown below in site config 
   ```yml
    editPost:
        URL: https://github.com/adityatelange/hugo-PaperMod/tree/exampleSite
        Text: "Edit"
   ```

layouts/partials/edit_post.html
adityatelange: @kdkasad can you explain why Edit post URL will contain a `mailto` link ?

kdkasad: > This is useful for providing the reader with an easy way to find the proper place to suggest changes.

If the site author wants the user to send an email to a certain address in order to request changes, they might want to use a mailto link there. For example:
```
mailto:webmaster@example.site?subject=suggested%20changes%20for%20example.site
```

adityatelange: The Edit link is focused here for open sourced websites which people can contribute to..

Syphdias: I am open to making the href more flexible and the `/content` suffix mandatory for the `editPostUrl`.
I am not that experienced enough with the framework to know if there would be cased where it would not be `content` but maybe a custom path. If there is such an option this could be accommodated with this change.

Also, I thought about if there might be options where you would not like to have `.File.Path` as the variable per post but something different and if this should be configurable.
For now I'd suggest leaving it as is: `editPostUrl` + `"/content/"` (or `"/"` depending on paragraph above) + `.File.Path`

Syphdias:
> 1. How about adding this to `post_meta` partial so that it would not display extra `·` when other meta params are empty

I moved the link to `post_meta` but I am unsure how to deal with `slice`. When I added `(slice (...))` around the `printf` I got stuck with an extra `·` separator but no text after it. I assume there was an extra element added to the `meta` list but an empty one.
To check if I made a mistake elsewhere I tried to simplify it:
```
{{ $scratch.Add "meta" (slice "foo") }} # didn't work
{{ $scratch.Add "meta" "bar" }}         # worked
```
I have no idea why `slice` is needed in the above cases but my case doen't seem to need it(?). Let me know if I did something wrong here.

> 2. It would be appropriate to keep `editPost` params with each other

Makes sense. Changed.

adityatelange:
@Syphdias I see edit button in `list` pages, we should find a way to hide this
image

Syphdias:
Hm, interesting, same for archives since it uses the `post_meta` partial as well.

I suggest going back to a separate partial like the translation again. This way it can easily be assured to only be used where appropriate, i.e. `single`.

I tired it analogues to the translations:
image
or
image

Also I found this dandy line which I could also use to only display the separator if there is date, auther, etc.:
```
{{- if or .Params.author $.Site.Params.author (.Param "ShowReadingTime") (not .Date.IsZero) }}&nbsp;|&nbsp;{{- end -}}
```

So a few questions:
- Which order do you prefer? meta, edit, translation or meta, translation, edit?
- Which separator? `|` or `·`?
  Also possible: `meta1 · meta2 · edit | translation` or
  `meta1 · meta2 | edit · translation` / `meta1 · meta2 | translation · edit` or
  `meta1 · meta2 · edit · translation` / `meta1 · meta2 · translation · edit`

mohnoor94:
`+1` nice feature!

Syphdias:
@Syphdias I implemented one of my suggestions from my last comment. I decided to go with `meta1 · meta2 | edit | translation`.
The reasoning behind it: Using `·` to softly separate meta data of post (date, reading time, author(s)) and `|` to strongly separate meta data from edit link as well as translation list.

I also picked up @kdkasad's suggestion to make the URL more flexible.

Would be great if you could give it another look.

Syphdias:
Hey @adityatelange, I was wondering if you could give it another look :slightly_smiling_face:

adityatelange:
> Hey @adityatelange, I was wondering if you could give it another look

Yes

adityatelange (commented):
Everything looks good to me :) , just can we move `Edit` link to end of `post-meta` ? 

---

Also maybe we should add a param named `NoGit` so that @kdkasad's previous [comment](https://github.com/adityatelange/hugo-PaperMod/pull/278#discussion_r584465422) could be taken into consideration so that a `mailto` link could be added for non-git edit requests

As if `NonGit = true` we could only link `URL` and not the path with it

Syphdias:
> Everything looks good to me :) , just can we move `Edit` link to end of `post-meta` ? Also how about opening links in new tab (don't forget to add `rel="noopener noreferrer"`)?

This is three things:
* "end of `post-meta`": The link is already _after_ the `post-meta` So I think you mean "after the translation"? Like this? `meta1 · meta2 | translation | edit` => need feedback if interpretation was correct
* `rel="noopener noreferrer"`: Frankly, I am not quite sure what these mean but it sounds SEO related and seems to make sense here. So I added it. => done
* Opening in new tap. The `rel` options don't do that, so you probably mean `target="_blank"` unless I am missing something. I am a big hater of the option because it gets overused and I use middle mouse button or <kbd>ctrl</kbd>, if I want to open in a new tab (I even use [Death To _blank](https://chrome.google.com/webstore/detail/death-to-blank/gneobebnilffgkejpfhlgkmpkipgbcno)) **but** here I think it makes sense. Also added. => feedback needed if this was wrong.

How wrong did I interpret this?


> Also maybe we should add a param named `NoGit` so that @kdkasad's previous [comment](https://github.com/adityatelange/hugo-PaperMod/pull/278#discussion_r584465422) could be taken into consideration so that a `mailto` link could be added for non-git edit requests
> 
> As if `NonGit = true` we could only link `URL` and not the path with it

`mailto://` already works:
With `URL: "mailto://mail@example.com?subject=Suggesting changes for "` you get:
```html
<a href="mailto://mail@example.com?subject=Suggesting%20changes%20for%20/posts/my-first-post.md" rel="noopener noreferrer" target="_blank">Edit</a>
```

So I would suggest adding another option, if you do not want the page. Something like `editPost.appendFilePath: false`. I'm also open to defaulting to `false` and having to manually set it to `true` if you want to do something like the initial idea.
=> needs feedback about idea, variable name, and default
 
> Also I have doubt with usage of `.Param` with string values, empty strings can make issues somehow..
> refer: https://gohugo.io/functions/param/ and [gohugoio/hugo#3366](https://github.com/gohugoio/hugo/issues/3366)
> 
> in that case `{{ .Params.editPost.URL | default .Site.Params.editPost.URL }}` should be use imo.

=> done, I hope I got everything. Question though: There is `$.Site.Params.author` – should I use the `$` prefix for `$.Site.Params.editPost.URL`?

adityatelange:
> This is three things:
> 
> * "end of `post-meta`": The link is already _after_ the `post-meta` So I think you mean "after the translation"? Like this? `meta1 · meta2 | translation | edit` => need feedback if interpretation was correct

End of post-meta means at the last position in the class `post-meta` 

> * `rel="noopener noreferrer"`: Frankly, I am not quite sure what these mean but it sounds SEO related and seems to make sense here. So I added it. => done
> * Opening in new tap. The `rel` options don't do that, so you probably mean `target="_blank"` unless I am missing something. I am a big hater of the option because it gets overused and I use middle mouse button or ctrl, if I want to open in a new tab (I even use [Death To _blank](https://chrome.google.com/webstore/detail/death-to-blank/gneobebnilffgkejpfhlgkmpkipgbcno)) **but** here I think it makes sense. Also added. => feedback needed if this was wrong.

Opening in new tab `target="_blank"` is a necessary code so I didn't mention that but the `rel` one.

> How wrong did I interpret this?

No you didn't. 

> So I would suggest adding another option, if you do not want the page. Something like `editPost.appendFilePath: false`. I'm also open to defaulting to `false` and having to manually set it to `true` if you want to do something like the initial idea.
> => needs feedback about idea, variable name, and default

Okay `editPost.appendFilePath` is fine, should be defaulting to `false` as most people will use git providers link

> > Also I have doubt with usage of `.Param` with string values, empty strings can make issues somehow..
> > refer: https://gohugo.io/functions/param/ and [gohugoio/hugo#3366](https://github.com/gohugoio/hugo/issues/3366)
> > in that case `{{ .Params.editPost.URL | default .Site.Params.editPost.URL }}` should be use imo.
> 
> => done, I hope I got everything. Question though: There is `$.Site.Params.author` – should I use the `$` prefix for `$.Site.Params.editPost.URL`? 

If it works no need to (info here https://gohugo.io/templates/introduction/#2-use--to-access-the-global-context)

adityatelange:
> Where is documentations supposed to go?
> 
> * `README.md` on branch  `master`?
> * Wiki of this repo?
> * `content/posts/papermod/papermod-variables.md` on branch `exampleSite`
> * Multiple places? (<-wrong answer)

Wiki of this repo. I clone the same to `exampleSite`, and not necessary it is up to date.
Yes it is multiple places
It just made that way so that new users may get an idea of stuff added into without visiting the wiki.

Syphdias:
> End of post-meta means at the last position in the class `post-meta`

`post-meta` is used in `single` (where we want the translations and edit), `search` and `archives`. If I put the edit button back into `post-meta` it will show up in two extra places where it is unwanted. I tried to differentiate between but I did not find a way.
I also think it is cleaner and more concise to have a separate partial. Like translation it isn't really meta data but data that ties multiple places together. In case of translation it connects two posts. In the case of edit it connects to a way to edit the article.
Is there any reason you want it in the `post-meta` file.

> Okay `editPost.appendFilePath` is fine, should be defaulting to `false` as most people will use git providers link

But wouldn't you want to append the file path if you used a git provider link?


> If it works no need to (info here https://gohugo.io/templates/introduction/#2-use--to-access-the-global-context)

Ah, thank you!

> Wiki of this repo. I clone the same to `exampleSite`, and not necessary it is up to date.

Ah, so the wiki is the complete one and `exampleSite` is a snapshot in time of the wiki.

Syphdias:
>> Okay editPost.appendFilePath is fine, should be defaulting to false as most people will use git providers link
>
> But wouldn't you want to append the file path if you used a git provider link?

I implemented `editPost.appendFilePath`. Setting `editPost.appendFilePath: true` is now needed of you want to use the file path as you suggested.

adityatelange (approved):

adityatelange:
@Syphdias Nice Work ! :+1:  Sorry for keeping this PR for too long.
Also I'll update the docs later on.

Syphdias:
> Also I'll update the docs later on.

I couldn't edit the wiki directly, so I cloned the `.wiki` repo and pushed it to separate repo: https://github.com/Syphdias/hugo-PaperMod-wiki

Feel free to change what I have written. Maybe it saves you some time.

openact:
appendFilePath appends the file name with backword slash.  
So link becomes https://github.com/....../content/posts\readme.md
Am I missing something? I am using vscode on windows.

Syphdias:
Oh, sorry about that.

> Am I missing something? I am using vscode on windows.

VSCode has nothing to do with it. Apparently `.File.Path` is dependent on the operating system which is not properly documented and I did not know – it is for `.File.Path`: https://gohugo.io/variables/files/

Quick and dirty fix below. Warning: Breaks other stuff like `\` in a filename.

```diff
diff --git a/layouts/partials/edit_post.html b/layouts/partials/edit_post.html
index 21a33eb5..925b2f8c 100644
--- a/layouts/partials/edit_post.html
+++ b/layouts/partials/edit_post.html
@@ -1,6 +1,7 @@
+{{- $fileUrlPath := path.Join (split .File.Path "\\") }}
 {{- if or .Params.editPost.URL .Site.Params.editPost.URL -}}
 {{- if or .Params.author $.Site.Params.author (.Param "ShowReadingTime") (not .Date.IsZero) .IsTranslated }}|&nbsp;{{- end -}}
-<a href="{{ .Params.editPost.URL | default .Site.Params.editPost.URL }}{{ if .Params.editPost.appendFilePath | default ( .Site.Params.editPost.appendFilePath | default false ) }}/{{ .File.Path }}{{ end }}" rel="noopener noreferrer" target="_blank">
+<a href="{{ .Params.editPost.URL | default .Site.Params.editPost.URL }}{{ if .Params.editPost.appendFilePath | default ( .Site.Params.editPost.appendFilePath | default false ) }}/{{ $fileUrlPath }}{{ end }}" rel="noopener noreferrer" target="_blank">
     {{- .Params.editPost.Text | default (.Site.Params.editPost.Text | default (i18n "edit_post" | default "Edit") ) -}}
 </a>
 {{- end }}
```

I guess the right way would be to use `path.Split` (in a loop) and the join with `/`. Also there is a weird note for [`path.Join`](https://gohugo.io/functions/path.join/):
> Note: All path elements on Windows are converted to slash ('/') separators.

Does that mean a this would convert a path? ``{{- $fileUrlPath := path.Join .File.Path }}` to unix style? Could you give that a try, @openact? I don't have a Winowds easily accessible right now.

openact:
yes.. i checked this.. @Syphdias .. this fix is working in windows..

Syphdias:
@openact The patch or the `path.Join .File.Path`?

openact:
I tried patch.. I did not get what to do in `path.join .File.Path`.. can you elaborate more? 
I do not have much knowledge in this..

Syphdias:
Let's say you have these file:
```
content/posts
└── bla foo
    ├── baz\qux.md
    └── quux.md
```
* Linux: `bla foo/baz\qux.md`
  Windows will probably have trouble with the backslash in the file name.
* Linux: `bla foo/quux.md`
  Windows: `bla foo\quux.md`

The patch does something simple: Take the `.File.Path` and split it at `\` and the join it again. If there is no `\` like in `bla foo/quux.md` while on Linux, no harm done.
But if you look at `bla foo/baz\qux.md`, it will get messed up into `bla foo/baz/qux.md` which is wrong.

`.File.Path` yields differnet results for WIndows and Linux, but the function `path.Join` is always Linux style apparently. So my idea was to use this property to convert a windows path into a Linux path. My (probably wrong) hope was that you don't have to split `.File.Path` first but directly use it with `path.Join`.
If you want to try it, replace `path.Join (split .File.Path "\\")` from the patch with `path.Join .File.Path`. I don't expect it to work but would be nice if it did.

If it does not work, the proper way to do this would be to use `path.Split` (OS aware) multiple times for each directory in `.File.Path` and the combine them again with `path.Join`.

Syphdias:
> Edit: probably `path.Join (path.Split .File.Path)` not `path.Join .File.Path`

Nope, I was right the first time...
`path.Join .File.Path` oddly converts into URL/Linux style paths. I'll open a PR...

I might be wrong, but I think this OS-independent behavior is intentional because it look like hugo uses this function internally to build URLs. So this should not break in future releases. \*knock on wood\*

PS: Fun fact: I can't normally git checkout my current test repo HEAD since it includes a file with `\` in the name. :joy:

Syphdias:
@openact, checkout #333 if it fixes your problem.

PR #291, opened 2021-03-05
Add support for `rtl` and `ltr` shortcodes

mohnoor94:
This is to support `rtl` and `ltr` short-codes.

### Use Case:
You may need to insert some contents in a different direction compared to your main site direction so that these shortcodes will come in handy.

### Important notes:
- These shortcodes will support nested shortcodes by default so that they won't `markdownify` your contents!
- If you want to `markdownify` (support `md` in) your contents, you can set the `md` parameter to `true`.

### Examples:
```
{{< rtl >}}
{{< blockquote >}}

مرحبا بك

{{< /blockquote >}}
{{< /rtl >}}
```
This will result in a `blockquote` in `rtl`.

---

```
{{< rtl md=true >}}

> مرحبا بك

{{< /rtl >}}
```
This will result in `rtl` contents which rendered using `md`.

---

```
{{< rtl true >}}

> مرحبا بك

{{< /rtl }}
```
This will result in `rtl` contents which rendered using `md` (same example above, but with positional argument instead of a named argument in the content file).

---

The `ltr` shortcode could be used in the same way but in the context of an `rtl` website.

### DIFF
--- a/layouts/_default/list.html
+++ b/layouts/_default/list.html
@@ -49,7 +49,7 @@
       {{- if .Draft }}<div class="entry-isdraft"><sup>&nbsp;&nbsp;[draft]</sup></div>{{- end }}
     </h2>
   </header>
-  {{- if (ne .Site.Params.hideSummary true)}}
+  {{- if (ne (.Param "hideSummary") true)}}
   <section class="entry-content">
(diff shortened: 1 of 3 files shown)


mohnoor94:
Replaced by #293

PR #292, opened 2021-03-06
Allow selective summary hiding in front matter

mohnoor94:
The previous style of getting params does not allow hiding post summary from the listing page in the front matter. With this change, the below can be done to hide the summary for a specific post or page. This change mirrors how `ShowBreadCrumbs` already works.

```
---
title: "Test Post"
hideSummary: true
---

post contents...
```


adityatelange:
What is the use case here ?

mohnoor94:
When you set the `summary` var to null or empty it shows an annoying three dots without anything before them in the home/blog pages. So this will allow you to hide it completely (while still providing a summary in case you want to show it later,  just flip or remove the new option).

adityatelange (approved):

PR #302, opened 2021-03-11
Add hideDescription param

tmakowski:
This change makes showing the description optional. With hideDescription
param it is possible to provide a custom description for a <head>
without displaying the description on a given page.

### DIFF
--- a/layouts/_default/archives.html
+++ b/layouts/_default/archives.html
@@ -2,10 +2,12 @@
 
 <header class="page-header">
   <h1>{{ .Title }}</h1>
-  {{- if .Description }}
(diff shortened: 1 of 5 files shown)


adityatelange:
Just remove the `description` param from front-matter.

tmakowski:
It is not quite the same behaviour. With removed description it is set to truncated text before the `<!--more-->` tag. Having `hideDescription` allows setting custom description that is not shown under the post title.

adityatelange:
I didn't get that .. but there is a `.Summary` front-matter var you can use to show description in meta but not display it under post

PR #306, opened 2021-03-16
Add itch.io logo to socials list.

Icantjuddle:
Upstream source is here: https://itch.io/press-kit#logos/black


Icantjuddle:
Wondering if I should replace `color=#000` from the upstream with `fill="currentColor"`?

adityatelange (approved):

Icantjuddle:
Thank you!

PR #271, opened 2021-02-28
Add i18n support for Arabic

Muhammad-Saad-01:

PR #309, opened 2021-03-19
Minor adjustments in Arabic translations

bassel:


adityatelange:
I don't mind but if you (@bassel ) and @Muhammad-Saad-01 (#271) could discuss it about right or wrong it, it would be nicer ..

bassel:
Both version are valid in terms of meaning, it's just that this one is kind of more grammatically correct since it uses the dictionary versions of the words. What do you think @Muhammad-Saad-01 ?

mohnoor94:
+1 for this.

adityatelange:
Merging

PR #311, opened 2021-03-20
Updated head.html for SEO

joeldeleep:
Added site-wide keywords for SEO, this will enable to add keywords scoped to the entire website.
For example, if the site is a personal blog of david john , this will allow adding keywords in site params
```yaml
params:
   keywords: ["david","david john","david john blog"]

```


adityatelange:
added by https://github.com/adityatelange/hugo-PaperMod/commit/4878383dd4370c682f974570718b08e6104fc9c3

PR #326, opened 2021-03-25
Fix color of "Name" class in Code Highlighting with light background

JannikArndt:
When using the "chroma styles" classes (via `hugo gen chromastyles --style=github > assets/css/extended/syntax.css`), the `css` for _Names_ is just

```css
.chroma .n {  }
```

That means the text will be coloured by this part:
https://github.com/adityatelange/hugo-PaperMod/blob/b6af9eddee7f816b01a68836344b36d30de4a492/assets/css/post-single.css#L209-L218

For chroma styles with a _light_ background, this is not readable (basically white on white).

This fix uses the `--primary` colour, so black for the light and white for the dark mode.

### DIFF
--- a/assets/css/post-single.css
+++ b/assets/css/post-single.css
@@ -210,7 +210,7 @@
     display: block;
     margin: auto 0;
     padding: 10px;
-    color: rgba(255, 255, 255, .8);
+    color: var(--primary);
     background: 0 0;
     border-radius: 0;
     overflow-x: auto;


assets/css/post-single.css
```diff
@@ -210,7 +210,7 @@
     display: block;
     margin: auto 0;
     padding: 10px;
-    color: rgba(255, 255, 255, .8);
+    color: var(--primary);
```
kdkasad, line 213: This breaks the default text color of code blocks with the site's light theme. It's almost black on black.

**After:**
screenshot from example site after change

**Before:**
screenshot of example site before change

JannikArndt: Damn, yes, of course it is 🙈 My oversight was that dark/light for the _page_ is separate from dark/light for the _code_, and you can combine both, more or less arbitrarily. (Example:[ I change the code-bg to match the page-bg](https://www.jannikarndt.de/blog/2021/03/using_scalap_javap_and_the_scala_compiler/), which is why the change works _for me_). 
A generic "fix" would need a dedicated "code-text-color" variable in the `theme-vars.css`, so that it can be set for light and dark, together with the `--hljs-bg`, but separate from `--primary`.

I'm not sure if the use case justifies the additional complexity, or if it makes more sense to overwrite that _per user_. 

In any case, I don't have the testing experience to make this work, so I'll close the pr. Thank you for checking! 🙏

kdkasad: I see what you're trying to do now. Would it work to just include a line like the following at the top of your `syntax.css` file?
```css
.chroma {
    color: /* your color here */;
}
```
This should then be overridden by the later definitions for specific classes.

JannikArndt: Hm, for _some_ cases, yes. But not if the code-background changes with light/dark as well. Also that file is generated by `hugo`, so I wouldn't change it but rather do all customisation inside the `custom.css`.

PR #336, opened 2021-03-30
fixes #331

### LINKED ISSUE #331: [BUG] Google Search Console complains: "Unparsable structured data: Incorrect value type" (`itemListElement` starts with comma)
- Hugo Version [ >=0.74 expected]: v0.82.0
 - Theme Version [e.g. v4.0, master, or commit-id ]: master

Google Search Console complains about the `BreadcrumbList` in the `single.html` pages:

image

The `BreadcrumbList` > `itemListElement` starts with a comma:

```json
{
  "@context": "https://schema.org",
  "@type": "BreadcrumbList",
  "itemListElement": [, 
    {
      "@type": "ListItem",
      "position":  2 ,
      "name": "Blog",
      "item": "https://www.jannikarndt.de/blog/"
    }, 
    {
      "@type": "ListItem",
      "position":  3 ,
      "name": "Using scalap, javap and the scala compiler to understand the journey from code to bytecode",
      "item": "https://www.jannikarndt.de/blog/2021/03/using_scalap_javap_and_the_scala_compiler/"
    }
  ]
}
```

on https://www.jannikarndt.de/blog/2021/03/using_scalap_javap_and_the_scala_compiler/.

I compared that to the example site:

```json
{
    "@context": "https://schema.org",
    "@type": "BreadcrumbList",
    "itemListElement": [
        {
            "@type": "ListItem",
            "position": 1,
            "name": "Posts",
            "item": "https://adityatelange.github.io/hugo-PaperMod/posts/"
        },
        {
            "@type": "ListItem",
            "position": 2,
            "name": "Rich Content",
            "item": "https://adityatelange.github.io/hugo-PaperMod/posts/rich-content/"
        }
    ]
}
```

for https://adityatelange.github.io/hugo-PaperMod/posts/rich-content/.

Looking at the code in `schema_json.html`, the line `{{- if (and $index)}}, {{end }}` skips the first entry:

https://github.com/adityatelange/hugo-PaperMod/blob/0c0c6847e56e137664c1633e7f3e1d6fdc7f067a/layouts/partials/templates/schema_json.html#L33-L44

That makes sense, because `$bc_pg` is defined as 

https://github.com/adityatelange/hugo-PaperMod/blob/0c0c6847e56e137664c1633e7f3e1d6fdc7f067a/layouts/partials/templates/schema_json.html#L23

The difference here is that the "base" url for the example site is https://adityatelange.github.io/hugo-PaperMod/ whereas for my site it's https://www.jannikarndt.de/.

In fact, when running the site locally with `hugo server`, the issue does _not_ occur, because the base path then is `localhost:1313`:

```json
{
    "@context": "https://schema.org",
    "@type": "BreadcrumbList",
    "itemListElement": [
        {
            "@type": "ListItem",
            "position": 1,
            "name": "Blog",
            "item": "http://localhost:1313/blog/"
        },
        {
            "@type": "ListItem",
            "position": 2,
            "name": "Using scalap, javap and the scala compiler to understand the journey from code to bytecode",
            "item": "http://localhost:1313/blog/2021/03/using_scalap_javap_and_the_scala_compiler/"
        }
    ]
}
```

<br><br><br><br>
 
 
 
------

<br><br><br><br>
 
…… long thinking break …… 🤦 
<br><br><br><br>

------

<br><br><br><br>

Okay, so `hugo` sets the `baseURL` parameter automatically. And it sets it "correctly". _My_ `config.yaml` on the other hand defined it as

```yaml
baseURL: "https://www.jannikarndt.de"
```

Note the missing trailing slash.

Nothing else in `hugo` breaks if you don't write that (at least I've noticed nothing). Actually, the Sample `config.yaml` on https://adityatelange.github.io/hugo-PaperMod/posts/papermod/papermod-installation/#sample-configyml even specifies it _without_ the slash 😄 

image

(That makes me feel a little less stupid 😅)

I would propose

1. to change the sample `config.yaml`
2. to make the `$bc_list` more resilient to this config 😊 

I would have created a PR, but my go-template-skills are not up to that level. 

I also decided to file an issue despite having solved this for myself, because 1. it let's others find this and the solution and 2. there is a chance to limit user-frustration here 😉

adityatelange:


adityatelange:
@JannikArndt This should fix it :crossed_fingers:

JannikArndt:
Yes, looks good 👏 Thank you!

adityatelange:
Didn't really work ... Output is glitched see here - https://github.com/adityatelange/hugo-PaperMod/commit/2d28b6caec226bbe6188864ba23b994bd229589a

some sections missed, position incorrect for pages .. (I've removed the merged commit Force Pushed for now)
I am still skeptical about how this stuff works in hugo.

PR #342, opened 2021-04-02
remove .Site.Authors references

cryptic-code:
## About the Edits
This is a small edit to the Opengraph and Twitter Card templates to remove the `.Site.Authors` references which is an unimplemented feature and also absent in Hugo Documentation. `.Site.Authors` value cannot be set under any circumstances.

Until Hugo clarifies this further, this should fix the problem.

It solves this #340 

Other references of the issue:
- https://discourse.gohugo.io/t/how-do-i-setup-the-authors-key-in-the-config-file-for-twitter-creator-and-article-author-meta-tags/32102
- https://github.com/gohugoio/hugo/pull/8344
- https://github.com/gohugoio/hugo/issues/4458

## How is the issue solved?
Since `.Site.Authors` cannot be set, `twitter:creator` meta tag never renders. Same is the case with the `article:author` tag. The following changes fix the issue by getting the account handles from `.Site.Social` instead.

### DIFF
--- a/layouts/partials/templates/opengraph.html
+++ b/layouts/partials/templates/opengraph.html
@@ -30,13 +30,11 @@
 {{ with .PublishDate }}<meta property="article:published_time" content="{{ .Format $iso8601 }}" />{{ end }}
 {{ with .Lastmod }}<meta property="article:modified_time" content="{{ .Format $iso8601 }}" />{{ end }}
 
-{{- range .Site.Authors }}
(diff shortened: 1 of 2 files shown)


adityatelange:
I don't think we should do this, actually it becomes difficult to track upstream templates for changes they make and the changes we have here. So I'll wait until it is merged in hugo's repo.

cryptic-code:
Okay. That's cool.

PR #343, opened 2021-04-02
Move post nav links to a partial section

mohnoor94:
This will allow easier overrides for post nav links without the need of overriding the entire "single" template.


adityatelange (approved):
:+1:

PR #345, opened 2021-04-04
Add buttons to copy code block contents

### LINKED ISSUE #344: [Feature Request] code copy
Hello!
First of all, thank you for your theme.

I'm using it really well, but I'd like to add one more feature.
If you look at this [link](https://themes.gohugo.io//theme/hugo-clarity/post/markdown-syntax/#code-block-with-hugos-internal-highlight-shortcode), you can see an `copy` icon when you mouse over the code block.

Screen Shot 2021-04-03 at 11 03 32 PM

kdkasad:
Adds a clickable "copy" link in the top-right corner of each code block.

If available, uses the navigator.clipboard API. Falls back to selecting
the text and calling document.execCommand('copy') to copy text.

I mainly focused on functionality, not style, so let me know if you want it to look different.

I tested on both Chrome and Firefox on Linux, and it works using both the `navigator.clipboard` API and the fallback method in both browsers.

Closes #344

### DIFF
--- a/assets/css/common/main.css
+++ b/assets/css/common/main.css
@@ -52,15 +52,14 @@ div.highlight {
     display: none;
     position: absolute;
     top: 4px;
-    right: 16px;
-    color: var(--secondary);
+    right: 4px;
+    color: rgba(255, 255, 255, 0.8);
+    background: rgba(78, 78, 78, 0.8);
+    border-radius: var(--radius);
+    padding: 0 5px;
+    font-size: 14px;
 }
 
 div.highlight:hover .copy-code {
     display: block;
-}
-
-.copy-code:hover {
-    cursor: pointer;
-    text-decoration: underline;
 }
--- a/assets/css/common/main.css
+++ b/assets/css/common/main.css
@@ -41,5 +41,21 @@
 }
 
 code {
-    direction: ltr
+    direction: ltr;
+}
+
+div.highlight {
+    position: relative;
+}
+
+.copy-code {
+    position: absolute;
+    top: 4px;
+    right: 16px;
+    color: var(--secondary);
+}
+
+.copy-code:hover {
+    cursor: pointer;
+    text-decoration: underline;
 }
--- a/assets/css/common/main.css
+++ b/assets/css/common/main.css
@@ -49,10 +49,15 @@ div.highlight {
 }
 
 .copy-code {
+    display: none;
     position: absolute;
     top: 4px;
     right: 16px;
     color: var(--secondary);
+}
+
+div.highlight:hover .copy-code {
+    display: block;
 }
 
 .copy-code:hover {
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -104,12 +104,12 @@
 
         const copybutton = document.createElement('button');
         copybutton.classList.add('copy-code');
-            copybutton.innerText = '{{- i18n "copy" | default "copy" }}';
+        copybutton.innerText = '{{- i18n "code_copy" | default "copy" }}';
 
         function copyingDone() {
-            copybutton.innerText = '{{- i18n "copied" | default "copied!" }}';
+            copybutton.innerText = '{{- i18n "code_copied" | default "copied!" }}';
             setTimeout(() => {
-                copybutton.innerText = '{{- i18n "copy" | default "copy" }}';
+                copybutton.innerText = '{{- i18n "code_copy" | default "copy" }}';
             }, 2000);
         }
 
@@ -128,7 +128,7 @@
             try {
                 document.execCommand('copy');
                 copyingDone();
-            } catch (e) {};
+            } catch (e) { };
             selection.removeRange(range);
         });
 
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -96,3 +96,34 @@
 
 </script>
 {{- end }}
+
+{{- if (not .Site.Params.disableCodeCopy) }}
+<script>
+    document.querySelectorAll('pre > code').forEach((codeblock) => {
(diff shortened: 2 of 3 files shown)


adityatelange:
@kdkasad Code looks good to me, works as expected ! 
Can you enable maintainer edits? so that I can change few styling stuff :)

assets/css/common/main.css
```diff
@@ -41,5 +41,26 @@
...
+div.highlight {
+    position: relative;
+}
+
+.copy-code {
+    display: none;
+    position: absolute;
+    top: 4px;
+    right: 16px;
+    color: var(--secondary);
```
adityatelange, line 56: Style can be 

```
    right: 4px;
    color: rgba(255, 255, 255, 0.8);
    background: rgba(78, 78, 78, 0.8);
    border-radius: var(--radius);
    padding: 0 5px;
    font-size: 14px;
```

layouts/partials/footer.html
```diff
@@ -96,3 +96,34 @@
 
 </script>
 {{- end }}
+
+{{- if (not .Site.Params.disableCodeCopy) }}
+<script>
+    document.querySelectorAll('pre > code').forEach((codeblock) => {
```
adityatelange, line 102: Maybe we can add the following here and place `copyingDone(copybutton);` at `$2` and `$3`

```js
    function copyingDone(copybutton) {
        copybutton.innerText = 'copied!';
        setTimeout(function () {
            copybutton.innerText = 'copy';
        }, 2000);
    }
```

layouts/partials/footer.html
```diff
@@ -96,3 +96,34 @@
...
+        const container = codeblock.parentNode.parentNode;
+
+        const copybutton = document.createElement('div');
```
adityatelange, line 105: this could be done to hide `$1`
```diff
- const copybutton = document.createElement('div');
+ const copybutton = document.createElement('button');

File: assets/css/common/main.css [hunk 4]
```diff
@@ -41,5 +41,26 @@
...
+}
+
+div.highlight:hover .copy-code {
+    display: block;
+}
+
+.copy-code:hover {
+    cursor: pointer;
+    text-decoration: underline;
```
adityatelange, line 65: `$1`

layouts/partials/footer.html
```diff
@@ -96,3 +96,34 @@
...
+        copybutton.classList.add('copy-code');
+        copybutton.innerHTML = 'copy';
```
adityatelange, line 107: Maybe we should use `innerText` rather than `innerHTML`

layouts/partials/footer.html
```diff
@@ -96,3 +96,34 @@
...
+                return;
+            }
+
+            const range = document.createRange();
+            range.selectNodeContents(codeblock);
+            const selection = window.getSelection();
+            selection.removeAllRanges();
+            selection.addRange(range);
+            try {
+                document.execCommand('copy');
```
adityatelange, line 121: `$3`

layouts/partials/footer.html
```diff
@@ -96,3 +96,34 @@
...
+    document.querySelectorAll('pre > code').forEach((codeblock) => {
+        const container = codeblock.parentNode.parentNode;
+
+        const copybutton = document.createElement('div');
+        copybutton.classList.add('copy-code');
+        copybutton.innerHTML = 'copy';
+
+        copybutton.addEventListener('click', (cb) => {
+            if ('clipboard' in navigator) {
+                navigator.clipboard.writeText(codeblock.textContent);
```
adityatelange, line 111: `$2`

layouts/partials/footer.html
kdkasad: It doesn't make the text underlined on hover like `$1` does, but I'll make the change anyways since it is really a button.

kdkasad:
> @kdkasad Code looks good to me, works as expected !
> Can you enable maintainer edits? so that I can change few styling stuff :)

Maintainer edits are now enabled.

kdkasad:
Also it might be a good idea to add translation keys for the "*copy*" and "*copied!*" texts.

layouts/partials/footer.html
adityatelange: @kdkasad Yes, it's a button we don't need it to be underlined

kdkasad: Oh right! I hadn't made the style changes to turn it into a button at that point. I'll fix that up.

adityatelange (approved):

adityatelange:
@kdkasad I just noticed this doesn't work with 

```yml
markup:
    highlight:
        lineNos: true
```

and code blocks not having a parent class `highlight` (ex. https://adityatelange.github.io/hugo-PaperMod/posts/markdown-syntax/)

kdkasad:
> @kdkasad I just noticed this doesn't work with
> 
> ```yaml
> markup:
>     highlight:
>         lineNos: true
> ```
> 
> and code blocks not having a parent class `highlight` (ex. https://adityatelange.github.io/hugo-PaperMod/posts/markdown-syntax/)

It works fine with line numbers enabled for me. However you're right about it not working for non-`highlight` code blocks (i.e. ones indented with 4 spaces).

adityatelange:
> > @kdkasad I just noticed this doesn't work with
> > ```yaml
> > markup:
> >     highlight:
> >         lineNos: true
> > ```
> > 
> > 
> > and code blocks not having a parent class `highlight` (ex. https://adityatelange.github.io/hugo-PaperMod/posts/markdown-syntax/)
> 
> It works fine with line numbers enabled for me. However you're right about it not working for non-`highlight` code blocks (i.e. ones indented with 4 spaces).

Yes you are correct, it works fine with `lineNos` enabled. Maybe I should've restarted hugo's server.

PR #348, opened 2021-04-09
feat: add keybase icon

makkes:


makkes:
oh wait, it already got added in the meantime. :-D nvm

PR #350, opened 2021-04-10
Create eo.yml

1000283:
Esperanto translation

### DIFF
--- a/i18n/eo.yml
+++ b/i18n/eo.yml
@@ -1,0 +1,25 @@
+- id: prev_page
+  translation: "« paĝo"
+
+- id: next_page
+  translation: "paĝo »"
+
+- id: read_time
+  translation:
+    one : "1 min"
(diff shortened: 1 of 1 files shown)


i18n/eo.yml
```diff
@@ -0,0 +1,25 @@
+- id: prev_page
+  translation: "« paĝo"
```
adityatelange, line 2: Can you remove these arrows? The template adds arrows, we don't need to add them in translation files

i18n/eo.yml
```diff
@@ -0,0 +1,25 @@
...
+
+- id: next_page
+  translation: "paĝo »"
```
adityatelange, line 5: Here too

i18n/eo.yml
1000283: Done

adityatelange (approved):
Thanks :+1:

PR #356, opened 2021-04-13
Add uk.yaml Ukrainian translation

geeqla:


adityatelange (commented):
There are few translation strings missing, can you add the remaining ones as from the English one below
https://github.com/adityatelange/hugo-PaperMod/blob/master/i18n/en.yaml

adityatelange (approved):
Thanks @geeqla :+1:

PR #359, opened 2021-04-14
Fix Error: add site dependencies: load resources: loading translations: "site/themes/hugo-PaperMod/i18n/eo.yml:1:1": failed to load translations: no unmarshaler registered for yml

cbrownstein:
Rename file to prevent this error from being received:

`Error: add site dependencies: load resources: loading translations: "site/themes/hugo-PaperMod/i18n/eo.yml:1:1": failed to load translations: no unmarshaler registered for yml`


adityatelange:
Hugo not detecting yml as yaml is something unexpected :/
Thanks for the fix !

adityatelange (approved):

PR #360, opened 2021-04-14
Fix Error: add site dependencies: load resources: loading templates: "site/themes/hugo-PaperMod/layouts/partials/head.html:79:1": parse failed: template: partials/head.html:79: unexpected unclosed action in command

cbrownstein:
Join lines to prevent the following error from being received:

`Error: add site dependencies: load resources: loading templates: "site/themes/hugo-PaperMod/layouts/partials/head.html:79:1": parse failed: template: partials/head.html:79: unexpected unclosed action in command`

NB: AFAICT, Hugo versions 0.81.0 and higher will not produce this error since those versions support newlines in template actions and commands. See: https://gohugo.io/news/0.81.0-relnotes/#newlines-in-template-actions-and-commands

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -76,8 +76,7 @@
 {{- $isHLJSdisabled := (.Site.Params.assets.disableHLJS | default .Params.disableHLJS ) }}
 {{- if (and (eq .Kind "page") (ne .Layout "archives") (ne .Layout "search") (not $isHLJSdisabled)) }}
 {{- if not .Site.Params.assets.disableFingerprinting }}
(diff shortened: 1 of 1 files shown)


adityatelange:
> NB: AFAICT, Hugo versions 0.81.0 and higher will not produce this error since those versions support newlines in template actions and commands. See: https://gohugo.io/news/0.81.0-relnotes/#newlines-in-template-actions-and-commands

Yes. Still should we merge this ?

cbrownstein:
> Yes. Still should we merge this ?

I say we should. Breaking the action across 2 lines doesn't seem to
serve any purpose. All it does is break backward compatibility.

PR #365, opened 2021-04-15
Fix copy code button to work on non-highlighted code blocks

### LINKED ISSUE #363: Copy code button doesn't appear for many code blocks
> @kdkasad I just noticed this doesn't work with
> 
> ```yaml
> markup:
>     highlight:
>         lineNos: true
> ```
> 
> and code blocks not having a parent class `highlight` (ex. https://adityatelange.github.io/hugo-PaperMod/posts/markdown-syntax/)

It works fine with line numbers enabled for me. However you're right about it not working for non-`highlight` code blocks (i.e. ones indented with 4 spaces).

**Edit:** it also doesn't work for code blocks created using backticks with no syntax/language specified.

_Originally posted by @kdkasad in https://github.com/adityatelange/hugo-PaperMod/pull/345#issuecomment-817197431_

kdkasad:
Fixes #363 

Turns out it was much easier than I expected.

### DIFF
--- a/assets/css/common/main.css
+++ b/assets/css/common/main.css
@@ -44,7 +44,7 @@ code {
     direction: ltr;
 }
 
-div.highlight {
+pre {
     position: relative;
 }
 
@@ -60,6 +60,6 @@ div.highlight {
     font-size: 14px;
 }
 
-div.highlight:hover .copy-code {
+pre:hover .copy-code {
     display: block;
 }
--- a/assets/css/common/main.css
+++ b/assets/css/common/main.css
@@ -58,6 +58,7 @@ pre {
     border-radius: var(--radius);
     padding: 0 5px;
     font-size: 14px;
+    font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;
 }
 
 pre:hover .copy-code {
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -88,7 +88,7 @@
 {{- if (and (eq .Kind "page") (ne .Layout "archives") (ne .Layout "search") (not .Site.Params.disableCodeCopy)) }}
 <script>
     document.querySelectorAll('pre > code').forEach((codeblock) => {
-        const container = codeblock.parentNode.parentNode;
+        const container = codeblock.parentNode;
 
         const copybutton = document.createElement('button');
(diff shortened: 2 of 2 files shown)


adityatelange:
@kdkasad This doesn't show the copy button as it is shown right now after we use 

```yml
markup:
    highlight:
        lineNos: true
```

image

kdkasad:
> @kdkasad This doesn't show the copy button as it is shown right now after we use
> 
> ```yaml
> markup:
>     highlight:
>         lineNos: true
> ```
> 
> image

For this first one, I can't reproduce the problem. Here's a screenshot of mine:
image


> image

For this one, I get the same problem, but I'm not sure how to fix it. I don't know how to position the button relative to what is visible rather than the whole `pre` element.

adityatelange:
> For this first one, I can't reproduce the problem. Here's a screenshot of mine:
> image

Oh, I forgot, `pygmentsUseClasses: true` is also necessary with

```yml
markup:
    highlight:
        lineNos: true
```

so that the code block is shown as a block with nums and code, instead of diff borders for nums and code.

adityatelange:
@kdkasad I think I have got the fix for #363  

Can you please test ? https://github.com/adityatelange/hugo-PaperMod/tree/fix/code-copy

kdkasad:
> @kdkasad I think I have got the fix for #363
> 
> Can you please test ? https://github.com/adityatelange/hugo-PaperMod/tree/fix/code-copy
> 
> And if possible, suggest any easy way for [9f5c739](https://github.com/adityatelange/hugo-PaperMod/commit/9f5c739b543c325d35561a94b457a4f58dac092b)

Looks good to me! I think 9f5c739b543c325d35561a94b457a4f58dac092b is fine; I was hoping for a way to automatically adapt to all cases, but I think checking for each case is easier.

There is an empty `else if` block [here](https://github.com/adityatelange/hugo-PaperMod/blob/9f5c739b543c325d35561a94b457a4f58dac092b/layouts/partials/footer.html#L125), but I think it's just to skip that case for the following `else if`s.

I'm closing this PR since your solution works and this one doesn't.

adityatelange:
> There is an empty `else if` block [here](https://github.com/adityatelange/hugo-PaperMod/blob/9f5c739b543c325d35561a94b457a4f58dac092b/layouts/partials/footer.html#L125), but I think it's just to skip that case for the following `else if`s.

Actually I see that hugo converts the js into optimized code which is minified. So it does the cleaning stuff for builds and removes the unnecessary conditions.

PR #370, opened 2021-04-19
Create vi.yaml

FrankFeenix:


i18n/vi.yaml
```diff
@@ -0,0 +1,25 @@
+- id: prev_page
+  translation: "Trang trước"
+
+- id: next_page
+  translation: "Trang tiếp theo"
+
+- id: read_time
+  translation:
+    one : "1 phút"
+    other: "{{ .Đếm }} phút"
```
adityatelange, line 10: Here `.Count` is a hugo variable, need not be translated!

Can you add `.Count` in place of `.Đếm`

FrankFeenix: Okay. Thank u

adityatelange (approved):
Thanks :+1:

PR #393, opened 2021-05-01
Fix xml+rss link header (#392)

JannikArndt:
Fixes #392 by not encoding `rss+xml`.

PR #396, opened 2021-05-02
Fix + in timezone in article:published_time and article:modified_time tags

JannikArndt:
I live in UTC+1:00, and noticed the same problem as with #393: The `+` gets html-encoded. So I applied the same fix here.

I also added some line breaks so that each tag is on its own line.


adityatelange:
@JannikArndt I still feel like this also should be reported to Hugo as this is an `internal template` and affects each and everyone using Hugo.

PR #397, opened 2021-05-03
Add Examples section to Readme.md

Boiarshinov:
First of all - thanks you for PaperMod theme!
I built my first static website with your theme and that was a great experience!

About PR:
While I setted up my site with Hugo PaperMod theme I had a lot of problems, because I was a newbie with Hugo.
So I learned a lot not only from wiki pages, but mostly from your example site. 
And I think that it can be helpful to everyone to have more examples of using this theme. 
Thats why I propose to add 'Examples' section to main page of this repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ The [demo](https://adityatelange.github.io/hugo-PaperMod/) includes a lot of doc
 -   Cover image for each post (with Responsive image support).
 -   Light/Dark theme (automatic theme switch a/c to browser theme and theme-switch button).
 -   SEO Friendly.
--   Muliple Author support.
+-   Multiple Author support.
 -   Search Page with Fuse.js
 -   Other Posts suggestion below a post
 -   BreadCrumb Navigation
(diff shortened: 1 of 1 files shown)


kdkasad:
I don't think adding a list of example sites to the repository is a good idea. It creates a lot of possibilities for rapid changes to add/remove links as people change their sites, which will just get in the way of development. Also, the official example site is associated with this theme, so it's almost guaranteed to be active as long as this theme is maintained.

That being said, I don't think it's a bad idea; I only think it shouldn't go in the repository. Maybe a separate page in the example site with links to other example sites would be better.

dittrichlucas (approved):

adityatelange:
> I don't think adding a list of example sites to the repository is a good idea. It creates a lot of possibilities for rapid changes to add/remove links as people change their sites, which will just get in the way of development. Also, the official example site is associated with this theme, so it's almost guaranteed to be active as long as this theme is maintained.
> 
> That being said, I don't think it's a bad idea; I only think it shouldn't go in the repository. Maybe a separate page in the example site with links to other example sites would be better.

@Boiarshinov I agree with this. Also, it would create an unnecessary sense of advertisement of blogs. 

 Moreover, ExampleSite does have almost complete documentation as well as it can be used as an example.

PR #404, opened 2021-05-08
i18n: complete French translations

DBLouis:


adityatelange (commented):
@DBLouis Thank you for these translations!
Can we also add the remaining strings (`edit_post`) as per the template here https://github.com/adityatelange/hugo-PaperMod/wiki/Translations#want-to-add-your-language-

adityatelange (approved):
Thanks @DBLouis

PR #407, opened 2021-05-09
feat: add slight delay in color mode transitions

sudo-kaizen:
Add slight delay in color mode transitions

[Here is a sample deployed](https://mystifying-euler-c42fa0.netlify.app/)

PR #408, opened 2021-05-09
Add missing semicolons in CSS files

kdkasad:
Many of the CSS files had no semicolon after the last rule in a block, which is proper syntax, but it's annoying when adding more rules. Since omitting the semicolons doesn't really have a benefit, I've decided to add them to ease editing the styles in the future.

(For example, they caused [this problem](https://github.com/adityatelange/hugo-PaperMod/pull/407#discussion_r628949132). Not a big deal, but still annoying and avoidable.)

### DIFF
--- a/assets/css/common/404.css
+++ b/assets/css/common/404.css
@@ -7,5 +7,5 @@
     justify-content: center;
     height: 80%;
     font-size: 160px;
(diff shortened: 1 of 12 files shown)


adityatelange (approved):
:+1:

PR #418, opened 2021-05-23
Make list unique to avoid duplicated results

### LINKED ISSUE #414: [BUG] Search results seem to be duplicated
**Describe the bug**

 - Device/Os: [e.g. Android 10]: MacOS Big Sur 11.2.3
 - Type: [e.g. Desktop/Mobile] Desktop
 - Browser and version [e.g. Chrome 86.0]: Firefox 78.10
 - Hugo Version [ >=0.82.0 expected]: v0.83.1
 - Theme Version [e.g. v4.0, master, or commit-id ]: master

I seem to be getting duplicate entries with search hits on my search page, see screenshot.

**EDIT:** This seems to only manifest within a development environment with `hugo server`. Once deployed with netlify, I'm not seeing the issue.

**Steps to reproduce the behavior:**
Enabled search following the Features page of papermod

Added

```
outputs:
  home:
    - HTML
    - RSS
    - JSON
```

Added `content/search.md` with the following contents:

```
---
title: "Search" # in any language you want
layout: "search" # is necessary
# url: "/archive"
# description: "Description for Search"
summary: "search"
---
```

**Expected behavior**:
Single entry results.

**Screenshots**
image

Happy to work with you to diagnose, please let me know if I can provide any more helpful info.

iranzo:
Applies uniq to search index to avoid duplicates in generated index.json

Closes #414

### DIFF
--- a/layouts/_default/index.json
+++ b/layouts/_default/index.json
@@ -1,5 +1,5 @@
 {{- $.Scratch.Add "index" slice -}}
-{{- range .Site.RegularPages -}}
+{{- range .Site.RegularPages | uniq -}}
     {{- if and (not .Params.searchHidden) (ne .Layout `archives`) (ne .Layout `search`) }}
     {{- $.Scratch.Add "index" (dict "title" .Title "content" .Plain "permalink" .Permalink "summary" .Summary) -}}
     {{- end }}


dittrichlucas (approved):

adityatelange:
@iranzo I have tested the patch but the problem continues in `hugo server` mode with or without `uniq`. The entries added are already unique imo as the file size and hash didn't change after using `uniq`!

iranzo:
Just in case, In my repo I've other changes, but are for adding admonitions, etc, but nothing else related with search

adityatelange:
> @adityatelange can you point me to your repository? The one I tested it worked with this change:
> 
> image
> 

Yeah, I can understand. The search duplicates doesn't appear everytime we run `hugo server`. It happens sometimes sometimes it doesn't. 

You can check if entries are duplicated by downloading /index.json and looking for file size or hash.

iranzo:
In my case, entries are not duplicate since I added that uniq, but in my index.json there's only some fields:

`dict_keys(['content', 'permalink', 'summary', 'title'])`

adityatelange:
Found out that the problem is not because of [scripts](https://github.com/adityatelange/hugo-PaperMod/blob/master/assets/js/fastsearch.js) for search and also not because of [index.json](https://github.com/adityatelange/hugo-PaperMod/blob/master/layouts/_default/index.json).

This happens due to the fast-reload template cache in Hugo, in the development environment. The `/index.json` values are duplicated because of this and not because they are not using `uniq`. 
As a result, I don't think there is a fix for this.

_Originally posted by @adityatelange in https://github.com/adityatelange/hugo-PaperMod/issues/414#issuecomment-861184221_
closing

PR #425, opened 2021-05-24
Allow to search posts based on tags

fahmifj:
Enable use of "tags" as keys/search pattern in search options.

Example Fuse configuration in config.yml:
```
fuseOpts:
        isCaseSensitive: false
        shouldSort: true
        location: 0
        distance: 1000
        threshold: 0.2
        minMatchCharLength: 2
        keys: ["title", "content", "summary",  "tags"]
```


dittrichlucas (approved):

adityatelange:
I do not think this is necessary as `tag` text would be something from  `title`  or `content` or `summary`. 
Hope you understand

PR #427, opened 2021-05-29
fix post_meta/reading_time

lucasew:
The template is expecting a string to generate the formated output string but it's receiving a number type (probably int).

This PR fixes this simple problem that cause bugs like the one below.

image


adityatelange (approved):
Thanks! :)

PR #433, opened 2021-06-06
 Add bn.yaml Bengali Translation

JonyBepary:


adityatelange (commented):
@JonyBepary Thank you for these translations!
Can we also add the remaining strings as per the template here https://github.com/adityatelange/hugo-PaperMod/wiki/Translations#want-to-add-your-language-

adityatelange (approved):

PR #436, opened 2021-06-11
Make the "PoweredBy" in the footer optional

pooyavahidi:
This change make the "PowerBy" section in the footer optional (control the visibility via params in the config file).

Not sure why there were no .gitignore file. I added one with just one entry for now to ignore the .DS_Store files (for MacOS)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+.DS_Store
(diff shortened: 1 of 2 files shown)


adityatelange:
Sorry, but I do want to display it :slightly_smiling_face: 

Is it too much to ask for a small "Powered by" string at end of the page?

pooyavahidi:
Sure, it doesn't remove it by default. It just give the option if you someone wants to do so. 
However, I do understand if you don't want to go ahead with this.

PR #438, opened 2021-06-15
js script

geronimo-iia:
Hello,

In this pull request, you will retrieve :

- Javascript resources are loaded at the end of the document, inside a script.html rather than in the head section. 

- I wanna introduce a way to customize js library:
   - I added a js extended folder to have a quick solution for few line of js (as css extended folder)
   - a script_extended partial template when we have to include library like jquery when its needed

- rewrote inner javascript in a dedicated papermod.js using buildin js feature of hugo. It's good for cache :)

Thanks a lot for your work!

Regards,
Jerome


adityatelange:
> * Javascript resources are loaded at the end of the document, inside a script.html rather than in the head section.
> * I wanna introduce a way to customize js library:
>   
>   * I added a js extended folder to have a quick solution for a few lines of js (as CSS extended folder)
>   * a script_extended partial template when we have to include library like jquery when it's needed
> * rewrote inner javascript in a dedicated papermod.js using built-in js feature of Hugo. It's good for cache :)

- Js resources are already `defer`ed so that they don't affect perf. It is as good as the scripts loaded in the end.
- If you are using `minify` from Hugo, the code won't be the same and is a lot smaller than it actually is!
- We already have `extend_head` and `extend_footer`, so IMO we don't need more script fetching templates like `extend_scripts`
- Also by current implementation we are avoiding scripts that are not used by the owner(disabled features)
- Also inlining scripts make the website faster to load, as we have a lot fewer scripts. The fewer connections made to the server for fetching resources the faster the website will load.

PR #462, opened 2021-07-07
Fix selection colors

marioortizmanero:

The selection colors look wrong with the dark theme:

image

So this fixes it by just inverting the colors of the selection.

No, only in my own site: https://github.com/marioortizmanero/nullderef.com/issues/17

## PR Checklist

### DIFF
--- a/assets/css/core/reset.css
+++ b/assets/css/core/reset.css
@@ -136,3 +136,8 @@ img {
 ::-webkit-scrollbar-thumb:hover {
     background: var(--secondary);
 }
+
+::selection {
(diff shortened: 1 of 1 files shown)


adityatelange:
image

This is how it looks on light mode for code selection.

---

The colour scheme for selection is your OS theme colour decided by system. IMO it should remain as it is.

marioortizmanero:
Yes, it's decided by the OS, but it doesn't know how to match the colors according to the theme of the site, this helps. The black/white selection colors is just an idea, you can propose whatever you like best.

PR #474, opened 2021-07-21
Add ability to hide the header from the profile page

wojtyniak:
Adds a new variable params.profileMode.hideHeader that can remove render.html from the profile page.

This allows user to hide header from the profile page. In my case the title of the website and the title on the profile page are the same, so I decided just to hide . 

It might be a better a better idea to make it somehow a page property, so any page could hide the header, but I decided to start with the simplest solution.

No, I just went forward and made the change.

## PR Checklist

(I'm not sure about this)

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -13,7 +13,9 @@
 {{- print " dark" }}
 {{- end -}}
 " id="top">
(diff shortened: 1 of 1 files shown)


wojtyniak:
This didn't get any traction and can be easily achieved with custom layouts.

Issue #426, opened 2021-05-29
Extra space causing issues in Cloudflare Pages build


adyanth:
I am trying to build a static site with cloudflare pages, and the build fails with:
```
<!--StartFragment-->
19:20:31.838 | Installing Hugo 0.54.0
-- | --
19:20:32.402 | Hugo Static Site Generator v0.54.0-B1A82C61A/extended linux/amd64 BuildDate: 2019-02-01T10:04:38Z
19:20:32.403 | Started restoring cached go cache
19:20:32.406 | Finished restoring cached go cache
19:20:32.434 | go version go1.14.4 linux/amd64
19:20:32.440 | go version go1.14.4 linux/amd64
19:20:32.442 | Installing missing commands
19:20:32.442 | Verify run directory
19:20:32.443 | Executing user command: hugo --gc --minify
19:20:32.489 | Error: "/opt/buildhome/repo/themes/PaperMod/layouts/partials/toc.html:71:1": parse failed: template: partials/toc.html:71: illegal number syntax: "-"
19:20:32.492 | Failed: build command exited with code: 255

<!--EndFragment-->

```
This seems to point to an extra space here:
https://github.com/adityatelange/hugo-PaperMod/blob/573e673efae17255fc46ae44670de9ce51faf9d8/layouts/partials/toc.html#L71

`"#{{- $cleanedID  -}}"` should be `"#{{- $cleanedID -}}"` ?

I do see that the version used by cloudflare is old (even thought I set the environment variable to 0.83.1), but that should not be causing the builds to fail, right?

adyanth:
Apparently the environment variables aren't saved until the site is saved. After that, in the new Hugo version, there weren't any errors. That space still seems out of place, but if nothing else, this issue can be closed.

adityatelange:
I never faced such error on newer `hugo` versions. I recommend using `hugo 0.82+` with this theme.

PR #488, opened 2021-08-02
Remove extraneous space in toc.html

lambdapioneer:

In some version of Hugo (e.g. the default one at Netlify) the extraneous space causes the following error:

```
8:12:29 PM: $ hugo
8:12:29 PM: Error: "/opt/build/repo/themes/PaperMod/layouts/partials/toc.html:71:1": parse failed: template: partials/toc.html:71: illegal number syntax: "-"
```

This PR removes the extraneous space.

Yes, in https://github.com/adityatelange/hugo-PaperMod/issues/426 - there the solution was to update the version of Hugo.

However, this is a low risk change and it makes the onboarding more pleasant for new users as they do not need extra configuration. Also, the error message does not guide them towards a possible solution (i.e. setting the Hugo version).

**Test plan**

I tested with unmodified version and observed the error shown above. Then I tested the modified version and the error no longer appeared.

## PR Checklist

### DIFF
--- a/layouts/partials/toc.html
+++ b/layouts/partials/toc.html
@@ -68,7 +68,7 @@
                 {{- end -}}
                 {{- end }}
                 <li>
-                    <a href="#{{- $cleanedID  -}}" aria-label="{{- $header | plainify -}}">{{- $header | safeHTML -}}</a>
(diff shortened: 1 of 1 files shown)


adityatelange (approved):
Thanks @lambdapioneer :+1:

Issue #149, opened 2020-12-23
Using profileMode UI on pages other than main


Allen-Hu:
As the title, is it possible to show the profileMode UI on something like an "About Me" page rather than the homepage?
Thanks in advance!

adityatelange:
Sorry no plans to this.. You may add it as a **shortcode** and put it on **about.md** page.

code for profile-mode page is here:
https://github.com/adityatelange/hugo-PaperMod/blob/0f994eee95811f5cf6099fb5897dd62be1e6cff2/layouts/partials/index_profile.html#L1-L21

PR #495, opened 2021-08-10
Fix profileMode imageUrl preloaded on every page

### LINKED ISSUE #494: [BUG] Using imageURL for profileMode preloads it on every page
Introduced in a86abf59947645d34c56ca31f6b94bcb99eeb750
Will be adding PR that resolves the issue shortly.

Device: Windows 10 Desktop (at least)
Browser: (New) Edge, Firefox, Chrome, latest stable versions as of 2021-08-09
Hugo Version (not relevant, but 0.85.0 through 0.87.0, at least)
PaperMod verision: 664299cc7cf96a3b410dfa7ec20a32378f4ec10b

**Steps to reproduce the behavior:**

Enable ``profileMode`` with a configuration such as (``config.toml``):

```toml
[params.profileMode]
    enabled = true
    title = "A Site Title"
    subtitle = "A subtitle"
    imageUrl= "/an-image.png"
    imageTitle = "An image alt text"
    imageWidth = 512
    imageHeight = 300
```
In (new) Edge, use 'developer tools' and notice the warning that /an-image.png is preloaded but not used within x seconds.

So, this does add the imageURL to the home page, unfortunately it also results in imageURL being preloaded on every page.

**Expected behavior**:
Image should only be preloaded on home page (where it used).

danielfdickinson:
Instead we now only preload on the home page.
Closes #494

Solves #494 by only preloading imageUrl on homepage.

Created issue and PR about same time. Will update PR if needed, but for such a simple issue/fix it is unlikely (in #149 @adityatelange indicated that profileMode is only for homepage).

## PR Checklist

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -59,7 +59,7 @@
 {{- end }}
 
 {{- with .Site.Params.profileMode }}
-{{- if and .enabled .imageUrl }}
+{{- if and .enabled .imageUrl $.IsHome }}
 <link rel="preload" href="{{ .imageUrl }}" as="image">
 {{- end }}
 {{- end }}


layouts/partials/head.html
```diff
@@ -59,7 +59,7 @@
 {{- end }}
 
 {{- with .Site.Params.profileMode }}
-{{- if and .enabled .imageUrl }}
+{{- if and .enabled .imageUrl .IsHome }}
```
adityatelange, line 62: Will this work as `IsHome` is not a part of `profileMode` obj.
Shouldn't this be `$.IsHome`?

danielfdickinson: Oops. I had originally done this as an if outside the with, but then got 'clever'. Will fix.

danielfdickinson: It turns out that it works as is as well as with $.IsHome (in testing with exampleSite with a modified config, both with and without either version of the fix) but it is 'safer' to use $.IsHome so I'll push that change in a moment.

adityatelange: I doubt that will work. As .IsHome doesn't exist the if condition will always be false, thus the preload won't work even on Home page.

adityatelange (approved):
Thanks 👍

PR #519, opened 2021-08-18
Add width and height attributes to cover images

igruenig:

When using cover images the layout shifts on initial load because the image dimensions can only be determined by the browser **after** loading the images.

I followed the recommendations from this [website](https://web.dev/optimize-cls/?utm_source=lighthouse&utm_medium=devtools#images-without-dimensions) linked to from Lighthouse in Chrome and added the width and height attributes to the cover images. From this the browser can determine the aspect ratio **before** loading any images.

Additionally, in the CSS file I have added the `height: auto` directive to the `.entry-cover img` selector. The height is then calculated based on the actual available width.

Fix for #499 in the Discussions

## PR Checklist

### DIFF
--- a/layouts/partials/cover.html
+++ b/layouts/partials/cover.html
@@ -19,8 +19,7 @@
             sizes="(min-width: 768px) 720px, 100vw" src="{{ $cover.Permalink }}" alt="{{ $alt }}" 
             width="{{ $cover.Width }}" height="{{ $cover.Height }}">
         {{- else }}{{/* Unprocessable image or responsive images disabled */}}
-        <img loading="lazy" src="{{ (path.Join .RelPermalink .Params.cover.image) | absURL }}" alt="{{ $alt }}" 
-            width="{{ $cover.Width }}" height="{{ $cover.Height }}">
+        <img loading="lazy" src="{{ (path.Join .RelPermalink .Params.cover.image) | absURL }}" alt="{{ $alt }}">
         {{- end }}
     {{- else }}{{/* For absolute urls and external links, no img processing here */}}
         {{- if $addLink }}<a href="{{ (.Params.cover.image) | absURL }}" target="_blank"
--- a/layouts/partials/cover.html
+++ b/layouts/partials/cover.html
@@ -23,7 +23,7 @@
             alt="{{ $alt }}"
             width="{{ $cover.Width }}"
             height="{{ $cover.Height }}"
-        />
+        >
         {{- else }}{{/* Unprocessable image or responsive images disabled */}}
         <img 
             loading="lazy"
@@ -31,12 +31,12 @@
             alt="{{ $alt }}"
             width="{{ $cover.Width }}"
             height="{{ $cover.Height }}"
-        />
+        >
         {{- end }}
     {{- else }}{{/* For absolute urls and external links, no img processing here */}}
         {{- if $addLink }}<a href="{{ (.Params.cover.image) | absURL }}" target="_blank"
             rel="noopener noreferrer">{{ end -}}
-            <img loading="lazy" src="{{ (.Params.cover.image) | absURL }}" alt="{{ $alt }}"/>
+            <img loading="lazy" src="{{ (.Params.cover.image) | absURL }}" alt="{{ $alt }}">
     {{- end }}
     {{- if $addLink }}</a>{{ end -}}
     {{/*  Display Caption  */}}
--- a/layouts/partials/cover.html
+++ b/layouts/partials/cover.html
@@ -16,9 +16,11 @@
                         {{ printf "%s %s" (($cover.Resize (printf "%sx" $size)).Permalink) (printf "%sw ," $size) -}}
                         {{ end }}
                     {{- end -}}{{$cover.Permalink }} {{printf "%dw" ($cover.Width)}}" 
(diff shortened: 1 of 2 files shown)


layouts/partials/cover.html
```diff
@@ -11,19 +11,32 @@
         {{- $processableFormats := (slice "jpg" "jpeg" "png" "tif" "bmp" "gif") }}
         {{- $prod := (hugo.IsProduction | or (eq .Site.Params.env "production")) }}
         {{- if (and (in $processableFormats $cover.MediaType.SubType) (ne .Site.Params.cover.responsiveImages false) (eq $prod true)) }}
-        <img loading="lazy" srcset="{{- range $size := $sizes -}}
-                        {{- if (ge $cover.Width $size) -}}
+        <img 
+            loading="lazy" 
+            srcset="{{- range $size := $sizes -}}
+                    {{- if (ge $cover.Width $size) -}}
```
danielfdickinson, line 17: This is an 'unrelated change'. It's not necessary and means that no upsizing occurs, where it did in the past. I know upsizing doesn't produce ideal images, but this should at least be a change that one opts into rather than having unexpectedly occur.

layouts/partials/cover.html
```diff
@@ -11,19 +11,32 @@
...
-                    {{- end -}}{{$cover.Permalink }} {{printf "%dw" ($cover.Width)}}"
-            sizes="(min-width: 768px) 720px, 100vw" src="{{ $cover.Permalink }}" alt="{{ $alt }}" />
+                    {{ end }}
+                {{- end -}}{{$cover.Permalink }} {{printf "%dw" ($cover.Width)}}"
+            sizes="(min-width: 768px) 720px, 100vw"
+            src="{{ $cover.Permalink }}"
+            alt="{{ $alt }}"
+            width="{{ $cover.Width }}"
+            height="{{ $cover.Height }}"
+        />
```
danielfdickinson, line 26: This is a minor nit, but if img tags elsewhere in source do not having the (optional) '/' then it shouldn't be added here either. I know some linters insist on it, but others insist on removing. The 'right' answer depends on the existing code base.

layouts/partials/cover.html
```diff
@@ -11,19 +11,32 @@
...
         {{- end }}
     {{- else }}{{/* For absolute urls and external links, no img processing here */}}
         {{- if $addLink }}<a href="{{ (.Params.cover.image) | absURL }}" target="_blank"
             rel="noopener noreferrer">{{ end -}}
-            <img loading="lazy" src="{{ (.Params.cover.image) | absURL }}" alt="{{ $alt }}">
+            <img loading="lazy" src="{{ (.Params.cover.image) | absURL }}" alt="{{ $alt }}"/>
```
danielfdickinson, line 39: Likewise here.

danielfdickinson (changes requested):
This are of course my opinions and may not be gospel truth.

layouts/partials/cover.html
igruenig: Thank you, missed those.

igruenig: I have no strong opinion on closing tags, I will remove them.

layouts/partials/cover.html
igruenig: Maybe I miss the point here, but I only reformatted that part. The size check was there before my change.

danielfdickinson: Oh, you are right, I missed that. I think the reformatting should be in a separate commit or PR, unless absolutely necessary

igruenig: Agreed. I reduced the changes to a minimum.

danielfdickinson: Thank you. Looks good to me now. Now to see what @adityatelange thinks...

igruenig: Thanks for reviewing!

danielfdickinson (commented):
Thank you for the changes. I'm not sure what I think of the formatting changes, mostly because I'm not sure what the maintainer prefers, but this looks good. Also, there is an issue #499 you might want to reference.

igruenig:
Thank you for the hint! I inserted the link in my initial comment.

layouts/partials/cover.html
```diff
@@ -15,10 +15,12 @@
...
                         {{ end }}
-                    {{- end -}}{{$cover.Permalink }} {{printf "%dw" ($cover.Width)}}"
-            sizes="(min-width: 768px) 720px, 100vw" src="{{ $cover.Permalink }}" alt="{{ $alt }}" />
+                    {{- end -}}{{$cover.Permalink }} {{printf "%dw" ($cover.Width)}}" 
+            sizes="(min-width: 768px) 720px, 100vw" src="{{ $cover.Permalink }}" alt="{{ $alt }}" 
+            width="{{ $cover.Width }}" height="{{ $cover.Height }}">
         {{- else }}{{/* Unprocessable image or responsive images disabled */}}
-        <img loading="lazy" src="{{ (path.Join .RelPermalink .Params.cover.image) | absURL }}" alt="{{ $alt }}">
+        <img loading="lazy" src="{{ (path.Join .RelPermalink .Params.cover.image) | absURL }}" alt="{{ $alt }}" 
+            width="{{ $cover.Width }}" height="{{ $cover.Height }}">
```
adityatelange, line 23: `Width` and `Height` won't exist for `Unprocessable image` right?

igruenig: Good point. I removed width and height for unprocessable images.

adityatelange (approved):

PR #497, opened 2021-08-10
Improve auto meta description for better SEO

danielfdickinson:
**What does this PR change? What problem does it solve?**

According to Bing SEO scan meta description should be between 15 and 160
characters long. In addition, Bing complains if there is no meta
description. We therefore make a best effort at a useful description
meets these requirements, unless the user has explicity provided one.

**Was the change discussed in an issue or in the Discussions before?**

I am sending as a _draft_ PR as I think it would be awkward to paste this into a discussion or issue, and a purpose of draft PRs is discussion of a proposal that is known to need work.

I would like to hear thoughts on this proposal. If you'd prefer to have the discussion in Discussions but leave the PR for referring to the code that works too.

## PR Checklist

- [X ] I have enabled [maintainer edits for this PR](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [X ] This change **does not** include any CDN resources/links.

PR #526, opened 2021-08-19
Add warning for bad meta description length (SEO)

danielfdickinson:

Warn on meta description <25 characters or >160 characters as this is
the range search engines prefer (according to Bing Webmaster Tools).

As I have added new strings I have made them translatable and provided the ``en`` translation. I haven't been able to figure out how to create a PR on the Wiki in order to provide an update to the list of translated strings there.

No, but #521 discusses a draft PR #497 which was determined to not be appropriate as it missed translation glue for new strings and automated changing the description, which was deemed inappropriate as the blogger / site author should write a good description themselves.

It is still useful to know if one has this issue (especially when first switching to PaperMod) at build time, therefore I have added this warning in the event the (SEO) issue exists.

Finally, I wonder if you would like me to update the formatting of the description block, since it's no longer embedded inside the meta content attribute?

## PR Checklist

- [X] This change adds/updates translations and I have used the [template present here](https://github.com/adityatelange/hugo-PaperMod/wiki/Translations#want-to-add-your-language-).

### DIFF
--- a/i18n/en.yaml
+++ b/i18n/en.yaml
@@ -23,3 +23,12 @@
 
 - id: code_copied
   translation: "copied!"
+
+- id: seo_colon
+  translation: "SEO:"
+
+- id: meta_description_less_than_25
+  translation: "meta description is less than 25 characters long"
+
+- id: meta_description_greater_than_160
+  translation: "meta description is greater than 160 characters long"
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -18,9 +18,17 @@
     {{- range $i, $e := .Params.keywords }}{{ if $i }}, {{ end }}{{ $e }}{{ end }} {{- else }}
     {{- range $i, $e := .Params.tags }}{{ if $i }}, {{ end }}{{ $e }}{{ end }} {{- end -}}" />
 {{- end }}
-<meta name="description" content="{{- with .Description }}{{ . }}{{- else }}{{- if .IsPage}}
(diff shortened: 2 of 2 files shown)


adityatelange:
Why would you like to provide warnings only for "description"?
There are a lot more things a user could miss to write to get "bad SEO"
For ex. Date, tags, or maybe forget to add a title?

**tl;dr**: Description is not everything a search engine needs to rank

danielfdickinson:
Well, I was planning to add more checks, but I didn't want to add unrelated changes. In addition, at least for myself, the ones that SEO checkers from Bing and Google complain about are description and title length (of things that I'd want to check for).
I'm happy to keep this in my own tree as a convenience function so that I have a straightforward way to make sure the description (and will be adding title) are within the recommended lengths, because I find it easy to go over the length (more so than under).

adityatelange:
> Well, I was planning to add more checks, but I didn't want to add unrelated changes. In addition, at least for myself, the ones that SEO checkers from Bing and Google complain about are description and title length (of things that I'd want to check for).
> I'm happy to keep this in my own tree as a convenience function so that I have a straightforward way to make sure the description (and will be adding title) are within the recommended lengths, because I find it easy to go over the length (more so than under).

The thing is we cannot make users tell each and everything that they are missing or doing it the wrong way. If you don't put an appropriate title WordPress is not gonna complain about it, rather it's the user's choice.
Hugo SSG is intended to be used by intermediate users and not non-technical people. Hope you understand.

danielfdickinson:
Not a problem. It's more a convenience thing for me that I thought might be useful for others.

PR #531, opened 2021-08-20
Make font-sizes relative to base font which is relative to screen

danielfdickinson:

* Making all sizes (except decoration such border widths and drop shadow) relative to the size of the base font means the theme can maintain it's "look & feel" while allow the theme's font sizes to be adjusted by changing only one size — the base font size.
* Changing the font size(s) if one is using fonts other than the default is important because some fonts are harder to read at the same pixel size.
* In order to achieve this, I first moved all sizes into a separate CSS directory from the rest of the CSS. This makes finding and changing font sizes and margins, padding, etc much easier (however I am not sure whether in the final version whether it makes sense to keep the separate -- I think an argument can be made either way)
* One I identified all the sizes that needed updated, I configure the base font size by setting the ``font-size`` of the ``html`` element, and making all other sizes (including the ``body`` font) relative (``rem``) to that font-size.
* If one uses a font-size of 16px for the base-font the relative positions and sizes do not change at all (a backend change that on the frontend it is a noop).
* To do that I made the ``body`` font ``1.125rem`` which is ``18px`` for a base font size of 16px (the current ``body`` font size)
* The reason I did it that way is the I intend the base font size (``16px``) to be the smallest font size on the  page (and set that font size via a variable called ``--min-font-size``). At present there is are a number of metadata fonts that break this rule (and I personally find those font sizes too small, even for metadata, and think that they are problematic from an accessibility perspective), so there is an open question of whether to rebase on those fonts instead (which means the base font size would actually be ``14px``, to maintain the overall size ratio, or whether to bump them to ``1rem`` instead of ``.875rem``.
* The final commit (which can be split into separate PR, if wished) automatically sets the base font size based on the screen size (it is basically a noop until a screen resolution of about ``1376px``[1]) because the minimum font size is set to 16px (default) (var(--min-font-size))
* That commit also sets a maximum pixel size of 32px (default) (var(--max-font-size))
* In addition the ``main-width`` and ``nav-width`` are changed to be ``75vw`` (75% of viewport size) and ``95vw`` respectively.
* The calculation of font size tries to keep the number of characters inside the ``main-width`` for the ``body`` font (not base font) at 55 characters (based on various typography recommendations, such as <https://baymard.com/blog/line-length-readability>, although that wasn't where I first learned of that recommendation (it's just the first hit that came up that had decent references).
* The CSS for base font size begins with a fallback for older browsers
* If clamp() and calc() are supported, they are preferred
* Followed by min(), max, and calc()
* The last modern method uses calc() and an ``@media`` query to cap the max font size.

[1] Note that under Windows 10 (at least) if one using a scaling factor >100% that the effective resolution in pixels is divided by that amount (so at 150% a 2K screen is in the ``1376px`` range, but the pixels are larger, in a way (but with smoothing rather than pixelization)).

A little in #530 and #496, but expect there will be discussion about the implementation, as there is a lot to consider.

I'm introducing this as a _draft_ PR as I am sure discussion will occur.

## PR Checklist

### DIFF
--- a/assets/css/common/404.css
+++ b/assets/css/common/404.css
@@ -1,11 +1,6 @@
 .not-found {
     position: absolute;
-    left: 0;
-    right: 0;
     display: flex;
     align-items: center;
     justify-content: center;
-    height: 80%;
-    font-size: 160px;
-    font-weight: 700;
 }
--- a/assets/css/common/archive.css
+++ b/assets/css/common/archive.css
@@ -1,44 +1,25 @@
-.archive-posts {
-    width: 100%;
-    font-size: 16px;
-}
-
-.archive-year {
-    margin-top: 40px;
-}
-
 .archive-year:not(:last-of-type) {
-    border-bottom: 2px solid var(--border);
+    border-bottom-style: 2px solid var(--border);
 }
 
 .archive-month {
     display: flex;
     align-items: flex-start;
-    padding: 10px 0;
-}
-
-.archive-month-header {
-    margin: 25px 0;
-    width: 200px;
 }
 
 .archive-month:not(:last-of-type) {
-    border-bottom: 1px solid var(--border);
+    border-bottom-style: 1px solid var(--border);
 }
 
 .archive-entry {
     position: relative;
-    padding: 5px;
-    margin: 10px 0;
 }
 
 .archive-entry-title {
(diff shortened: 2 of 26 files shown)


danielfdickinson:
Closing as I am splitting (and making some modifications).

PR #540, opened 2021-08-29
Add webp to processable formats

danielfdickinson:
It adds webp as cover image 'processable format' so that webp images can also be responsive.

No.
## PR Checklist

### DIFF
--- a/layouts/partials/cover.html
+++ b/layouts/partials/cover.html
@@ -8,7 +8,7 @@
         {{- if $addLink }}<a href="{{ (path.Join .RelPermalink .Params.cover.image) | absURL }}" target="_blank"
             rel="noopener noreferrer">{{ end -}}
         {{- $sizes := (slice "360" "480" "720" "1080" "1500") }}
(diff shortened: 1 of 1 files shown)


adityatelange:
> "Note that you need the extended version of Hugo to encode to WebP."
> ~ https://gohugo.io/news/0.83.0-relnotes/

PaperMod which in this case is free from the Extended Version of Hugo, which is why it is not included

danielfdickinson:
Ah, ok. Feel free to close then (or would it be okay to add it as a configurable item)?

danielfdickinson:
According to [Hugo Functions](https://gohugo.io/functions/hugo/) ``hugo.IsExtended`` was added in v0.83.0. To save time, I will update the PR using it, and if you don't want, you can close.

danielfdickinson:
Moved to draft as changes required that will require testing.

danielfdickinson:
That was far easier than expected. Do not this does bump the minimum required Hugo version to 0.83.0 (not part of this PR).

PR #541, opened 2021-08-29
Font spacing relative to base font

danielfdickinson:

* Making all sizes (except decoration such border widths and drop shadow) relative to the size of the base font means the theme can maintain it's "look & feel" while allowing the theme's font sizes to be adjusted by changing only one size — the base font size.
* Changing the font size(s) if one is using fonts other than the default is important because some fonts are harder to read at the same pixel size.
* In order to achieve this, I first moved all sizes into a separate CSS directory from the rest of the CSS. This makes finding and changing font sizes and margins, padding, etc much easier (however I am not sure whether in the final version whether it makes sense to keep the separate -- I think an argument can be made either way)
* One I identified all the sizes that needed updated, I configure the base font size by setting the ``font-size`` of the ``html`` element, and making all other sizes (including the ``body`` font) relative (``rem``) to that font-size.
* If one uses a font-size of 12px for the base-font the relative positions and sizes do not change (a backend change that on the frontend it is a no-op).
* The reason I did it that way is the I intend the base font size (``12px``) to be the smallest font size on the  page (and set that font size via a variable called ``--min-font-size``).

A little in #530 and I had a draft PR (#531) which received no comments, so I split this change out from that PR, and made rebased the relative sizes based on the smallest font in all CSS in the theme.

## PR Checklist

### DIFF
--- a/assets/css/common/404.css
+++ b/assets/css/common/404.css
@@ -1,11 +1,6 @@
 .not-found {
     position: absolute;
-    left: 0;
-    right: 0;
     display: flex;
     align-items: center;
     justify-content: center;
-    height: 80%;
-    font-size: 160px;
-    font-weight: 700;
 }
--- a/assets/css/common/archive.css
+++ b/assets/css/common/archive.css
@@ -1,12 +1,3 @@
-.archive-posts {
-    width: 100%;
-    font-size: 16px;
-}
-
-.archive-year {
-    margin-top: 40px;
-}
-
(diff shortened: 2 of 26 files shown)


danielfdickinson:
I have used exampleSite with this, but if there were more specific testing you would like, please let me know. Basing the font sizes on the smallest font results in fractions that don't translate well into decimal, and I am not sure of the best way to handle that.

PR #544, opened 2021-08-30
Add option to convert image resources to another format

danielfdickinson:
This adds converting images to a different, supported, format
(e.g. wepb).

This adds the option of converting images that are page resources (both cover and markdown images) to another format (e.g. you can make all images webp).

With Hugo 0.87.0 this works for the markdown render-image hook, but I do not know how far back one can go with Hugo versions. In the past page (and site params, through ``.Page.Site.params``) params may not have been available to render hooks.

The ``cover.html`` portion of this commit is pretty trivial. It just checks for the param (page or site)``convertImagesTo`` and uses the contents of that as an option string for Hugo image processing.

The render hook is  a little tricker.

We have to access resources and params through the ``.Page`` variable (and top-level ``.Site.Params``  or ``site.Params`` are not available).

We do so, and if the image is a page bundle resource (i.e. we are able to use Hugo image processing) then we convert the image as request.

In other case we maintain previous ``render-image.html`` behaviour.

#539 is a question on how to do this for ``cover.html`` and #542 includes a bit of render hook discussion, although it's main focus is the ``render-link.html`` render hook.

## PR Checklist


danielfdickinson:
This has been tested with Hugo Extended 0.87.0 but not with other versions of Hugo. With Hugo Extended 0.87.0 it works as expected/intended.

layouts/_default/_markup/render-image.html
```diff
@@ -1 +1,21 @@
...
+{{ $destination := .Destination }}
+{{- if not (strings.HasPrefix .Destination "http") -}}
+    {{- $image := (.Page.Resources.ByType "image").GetMatch .Destination -}}
+    {{- if $image }}
+        {{- with .Page -}}
+            {{- $imageOptions := "" -}}
+            {{- with .Params.convertimagesto -}}
+                {{- $imageOptions = printf "%dx%d %s" $image.Width $image.Height . -}}
+            {{- else -}}
+                {{- with .Site.Params.convertimagesto -}}
```
adityatelange, line 10: Are you sure `.Site.Params` can be called within render hooks?

adityatelange: OMG! Yes.

danielfdickinson: LOL! I feel better now 👍 

Thank you for looking at this. I'll look at your changes shortly (for my understanding; it's all about collaboration and learning together, right?)

layouts/_default/_markup/render-image.html
```diff
@@ -4,10 +4,10 @@
 {{- if $image }}
     {{- with .Page -}}
         {{- $imageOptions := "" -}}
-        {{- with .Params.convertimagesto -}}
+        {{- with .Params.convertImagesTo -}}
             {{- $imageOptions = printf "%dx%d %s" $image.Width $image.Height . -}}
         {{- else -}}
-            {{- with .Site.Params.convertimagesto -}}
+            {{- with .Site.Params.convertImagesTo -}}
```
danielfdickinson, line 10: I thought that using .Params.varable only worked with all lowercase; or did I miss that also being fixed (it used to be the case, and there was a note in the docs about it, but I can't find it now).

layouts/_default/_markup/render-image.html
danielfdickinson: Also note this is actually .Page.Site.Params (because of the with). Plain .Site.Params does not work.

layouts/_default/_markup/render-image.html
danielfdickinson: Oh, cool it works with camelCase now! (I spun up a test).

danielfdickinson:
I noticed something while merging your changes into my experimental tree: I missed converting the full-size image (original) to the ``convertImagesTo`` format. I've figured out the changes to fix that, but I wonder if this is desirable, or if the link to the full-size original should be kept?

Here is the relevant snippet from an experimental version that converts the link and original image:

```
    {{- if $cover -}}{{/* i.e it is present in page bundle */}}
        {{- if $addLink }}<a href="{{ ($cover.Resize (printf "%dx%d%s" $cover.Width $cover.Height $imageConversion)).RelPermalink }}" target="_blank"
            rel="noopener noreferrer">{{ end -}}
        {{- $sizes := (slice "90" "180" "360" "480" "720" "1080" "1500") }}
        {{- $processableFormats := (slice "jpg" "jpeg" "png" "tif" "bmp" "gif" "webp") }}
        {{- $prod := (hugo.IsProduction | or (eq .Site.Params.env "production")) }}
        {{- if (and (in $processableFormats $cover.MediaType.SubType) (ne .Site.Params.cover.responsiveImages false) (or (eq $prod true)  (eq $isThumbnail true))) }}
        <img loading="lazy" srcset="{{- range $size := $sizes -}}
                        {{ if ge $cover.Width $size }}
                        {{ printf "%s %s" (($cover.Resize (printf "%sx%s" $size $imageConversion)).RelPermalink) (printf "%sw ," $size) -}}
                        {{ end }}
                    {{- end -}}{{ ($cover.Resize (printf "%dx%d%s" $cover.Width $cover.Height $imageConversion)).RelPermalink }}"
            sizes="{{ if ne $isThumbnail true }}(min-width: 768px) 720px, 100vw{{ else }}((min-width: 2048px) and (max-width: 4096px)) 360px, ((min-width: 1024px) and (max-width: 2048px)) 180px, (min-width: 768px) 90px, 100vw{{ end }}" src="{{ ($cover.Resize (printf "%dx%d%s" $cover.Width $cover.Height $imageConversion)).RelPermalink }}" alt="{{ $alt }}"
            width="{{ $cover.Width }}" height="{{ $cover.Height }}">

```
I'd like to know your preferred behaviour here.

danielfdickinson:
I hope the full-size (original) image conversion is okay with you; I pushed a lightly tested commit that ensures that the full-size (original-image) is converted for cover images and not just the responsive images. It seems to work on my site and with the demo example site PR I submitted a few days ago (that has cover images on some posts).

danielfdickinson:
I think I should decouple the ``render-link`` from the ``cover`` changes; I've been discovering some 'wrinkles' to the ``render-link`` aspect. How is best to do that?

PR #548, opened 2021-09-01
Allow page-based overriding of copy code buttons

kdkasad:
Allows specifying "showCodeCopyButtons: true/false" in the front matter
of a page. If it is not specified, the value from the site-wide
configuration is used.

It allows enabling/disabling code copy buttons for a specific page while keeping the default value for the rest of the site.

## PR Checklist

### DIFF
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -82,7 +82,7 @@
 </script>
 {{- end }}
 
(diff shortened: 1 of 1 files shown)


adityatelange (changes requested):
You haven't verified that this works, right?

kdkasad:
> You haven't verified that this works, right?

I have. It works as expected. I would've made it a draft PR if I hadn't verified it.

PR #551, opened 2021-09-03
Fix button not allowed inside anchor in footer

### LINKED ISSUE #549: [BUG] <button> element not allowed inside anchor (<a>)
HTML validators complain (correctly) about the footer "top-link" which has a \<button> inside an anchor (\<a>) element.

As per [Mozilla Developer Network page for \<a> element allowed properties](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#properties) 'interactive elements' (including \<button>) are not allowed inside an anchor element (\<a>).

I am experimenting with replacing the \<button> with \<span> (which is allowed) and will hopefully soon have a PR available.

danielfdickinson:

HTML validators complain (correctly) about the footer "top-link" which has a \<button> inside an anchor (\<a>) element.

As per [Mozilla Developer Network page for \<a> element allowed properties](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#properties) 'interactive elements' (including \<button>) are not allowed inside an anchor element (\<a>).

I replace the ``<button>`` with a ``<span>`` and add an ``aria-hidden="true"`` to avoid screen readers repeating ``go to top``, which is already present as an ``aria-title`` on the anchor element.

Closes #549

It is currently in action on <https://www.princesandmadmen.ca>.

## PR Checklist

### DIFF
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -15,11 +15,11 @@
 
 {{- if (not .Site.Params.disableScrollToTop) }}
 <a href="#top" aria-label="go to top" title="Go to Top (Alt + G)">
-    <button class="top-link" id="top-link" type="button" accesskey="g">
+    <span class="top-link" id="top-link" accesskey="g" aria-hidden="true">
         <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 12 6" fill="currentColor">
             <path d="M12 6H0l6-6z" />
         </svg>
-    </button>
+    </span>
 </a>
 {{- end }}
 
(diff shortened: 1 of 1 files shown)


adityatelange (commented):
Or maybe we can remove `span` also, and put the attributes to `a` tag

danielfdickinson:
I suspect the CSS styling trick that creates the up arrow requires that there be a sub-element, but I will test.

danielfdickinson:
I made the requested change, and it works as expected for me (the access key still works, it looks right, hover is good, the icon is only displayed after scrolling down a bit, etc).

adityatelange (approved):
LGTM

PR #555, opened 2021-09-05
Use HUGO 0.87's date formatting feature with time.Format

Saxodwarf:

In the `post_meta.html` partial file, use `.Date | time.Format` instead of `.Date.Format` to format the date, to take advantage of HUGO's new localized date formatting feature when formatting the date in the meta informations.
This time.Format's behaviour was included in HUGO 0.87, as described in [the documentation](https://gohugo.io/functions/dateformat/).

No

## PR Checklist

### DIFF
--- a/layouts/partials/post_meta.html
+++ b/layouts/partials/post_meta.html
@@ -1,7 +1,7 @@
 {{- $scratch := newScratch }}
 
 {{- if not .Date.IsZero -}}
-{{- $scratch.Add "meta" (slice (.Date.Format (default "January 2, 2006" .Site.Params.DateFormat))) }}
+{{- $scratch.Add "meta" (slice (.Date | time.Format (default "January 2, 2006" .Site.Params.DateFormat))) }}
 {{- end }}
 
 {{- if (.Param "ShowReadingTime") -}}


RoneoOrg:
Hi @Saxodwarf ,

the same `Date.Format` you are changing here also appears in [rss.xml](https://github.com/adityatelange/hugo-PaperMod/blob/573e673efae17255fc46ae44670de9ce51faf9d8/layouts/_default/rss.xml#L39) and [post_meta.html](https://github.com/adityatelange/hugo-PaperMod/blob/537ec0038f9016f3833020521bcf82179a38da4e/layouts/partials/post_meta.html#L4)

Do you think these occurrences should be changed too?

Saxodwarf:
Hi @RoneoOrg,

I believe the `post_meta.html` file I'm editing in this pull request is the one you mentioned in your comment.

I saw the use of `Date.Format` in `rss.xml`, but there, the date format is fixed and respects [RFC 822](https://datatracker.ietf.org/doc/html/rfc822#section-5.1) as requested in the RSS specification, so there is no need to use the new feature for date localization.

For the sake of coherence, we could edit `rss.xml` to use `time.Format`, but imho it is not useful. What do you think ?

RoneoOrg:
Hi @Saxodwarf 

> I believe the `post_meta.html` file I'm editing in this pull request is the one you mentioned in your comment.

You're right, I scanned the whole repo _in its current version!_

> I saw the use of `Date.Format` in `rss.xml`, but there, the date format is fixed and respects [RFC 822](https://datatracker.ietf.org/doc/html/rfc822#section-5.1) as requested in the RSS specification, so there is no need to use the new feature for date localization.
> 
> For the sake of coherence, we could edit `rss.xml` to use `time.Format`, but imho it is not useful. What do you think ?

As `.Date.Format` is far from being deprecated, let's stick with the current implementation
"_If it's not broken don't fix it_"

Thanks for your reference of the RFC, enlightening!

JoCat:
> As `.Date.Format` is far from being deprecated, let's stick with the current implementation "_If it's not broken don't fix it_"

This is cool, but doesn't work as expected on non-English sites.

How should it work:
image

How does it work:
image

How I solved the problem:
1) Use time.Format (https://gohugo.io/functions/dateformat/#datetime-formatting-layouts)
As in the pull request
2) Use .Date.Local (https://gohugo.io/functions/format/#use-local-and-utc)
`{{- $scratch.Add "meta" (slice (dateFormat (default "January 2, 2006" .Site.Params.DateFormat) .Date.Local)) }}`

Saxodwarf:
Hi all,
I just resolved a merge conflict with the last version of `master`.
I hope this can be merged soon :)

nielsbrakel (approved):

nielsbrakel:
Tested it with my multilingual site and works perfect!

adityatelange:
@Saxodwarf @nielsbrakel  Does this have backwards compatibility? 

Would this change work with Hugo 0.83 or do we need to update the min Hugo version also?

nielsbrakel:
@adityatelange This is a new feature since 0.87 so I suspect it is not backwards compatible. 

I would suggest raising the min supported version in the next major release.

Saxodwarf:
> Would this change work with Hugo 0.83 or do we need to update the min Hugo version also?

@adityatelange Yes. Although the internationalized date format feature is new in 0.87, the date.time function is not.
I just tested with Hugo 0.83, and the date is correctly displayed if we do not use the new layout strings.

If we keep the current, non-internationalized default, as I did in the PR, the change should be backward-compatible, and those of us who want to use the new feature can just put a `DateFormat: ":date_long"`parameter, or something similar in their configuration file.
So my take is that we do not need to increase the minimum Hugo version.

adityatelange (approved):

PR #559, opened 2021-09-08
Fix schema_json when baseURL omits trailing slash

### LINKED ISSUE #558: [BUG] structured data broken when baseURL has no trailing slash (/)
**Describe the bug**

Google Search Console reports an error in structured data when breadcrumbs are enabled and the baseURL is configured without a trailing slash (/), for example.

```yaml
baseURL: https://www.princesandmadmen.ca
```
This results in the following in \<head>

```html
<html>
<body>
<script type=application/ld+json>
{
  "@context": "https://schema.org",
  "@type": "BreadcrumbList",
  "itemListElement": [, 
  {
  "@type": "ListItem",
  "position":  2 ,
  "name": "Blog",
  "item": "https://www.princesandmadmen.ca/post/"
  }
 ]
}
</script>
```
I will have a PR (momentarily) that resolves the issue, but another option is just to document that a trailing slash is required.

danielfdickinson:

Google Search Console warns of syntax errors in structured data (the
breadcrumb trail) when breadcrumbs are enable and the .Site.BaseURL
is configured without a trailing slash (/).

This commit fixes the schema_json partial by temporarily adding the
omitted slash (/) so that the breadcrumb splitting works as expected.

Closes #558 

## PR Checklist

Not sure on the last task: I don't see a ``schema_json`` in Hugo repo, but I do see a ``schema``.

### DIFF
--- a/layouts/partials/templates/schema_json.html
+++ b/layouts/partials/templates/schema_json.html
@@ -18,7 +18,11 @@
 </script>
 {{- else if (or .IsPage .IsSection) }}
 {{/* BreadcrumbList */}}
-{{- $url := replace .Parent.Permalink ( printf "%s" .Site.BaseURL) "" }}
+{{- $base_url := urls.Parse .Site.BaseURL -}}
+{{- if eq $base_url.Path "" -}}
+{{- $base_url = urls.Parse (printf "%s/" .Site.BaseURL) -}}
+{{- end -}}
+{{- $url := replace .Parent.Permalink ( printf "%s" ($base_url | string)) "" }}
 {{- $lang_url := strings.TrimPrefix ( printf "%s/" .Lang) $url }}
 {{- $bc_list := (split $lang_url "/")}}
 


danielfdickinson:
Since the issue is there and has the info is it okay to go ahead and close this PR (with apologies for forgetting to check the closed issues), or would you rather keep this open with ``wontfix`` so that others don't make the same mistake? (Alternatively, you could add another task check to the PR template such as the one below). I'm not sure about others, but I do carefully go over that task list).

```markdown
[ ] I have checked closed issues and pull requests before submitting.
```

adityatelange:
> Since the issue is there and has the info is it okay to go ahead and close this PR (with apologies for forgetting to check the closed issues), or would you rather keep this open with ``wontfix`` so that others don't make the same mistake? (Alternatively, you could add another task check to the PR template such as the one below). I'm not sure about others, but I do carefully go over that task list).
> 
> ```markdown
> [ ] I have checked closed issues and pull requests before submitting.
> ```

Yes I'll close this. 
About the task list, we already have `Was the change discussed in an issue or in the Discussions before?` in the template. It covers the task list you have mentioned.

I think we should mention this in docs about this `/` thing in `baseURL`. But again it is with the `/` in sample config.yml part in docs.

danielfdickinson:
Could ``Was the change discussed in an issue or in the Discussions before?`` be changed to ``Was the change discussed in an issue (including closed) or in the Discussions before?``

At least for me, mentioning closed issues prompts my memory, especially when I'm tired or such.

Issue #557, opened 2021-09-06
[BUG] Emojis affected by theme opacity values


aormsby:
**Describe the bug**
Because the theme's colors are set using opacity instead of actual color values, they interfere with elements other than text. For example, dark elements styled by `entry-content` use the `--secondary` color of `rgba(255, 255, 255, 0.56)`. The text reads fine as a grey color against the black background, but if you add an emoji into your content it will also take on the 0.56 opacity value and look like it's fading away.

**Expected behavior**:
I would expect things like emojis to appear at full opacity. This could be fixed by using actual color values in the theme as opposed to values determined only by opacity.

For example:
`rgba(255, 255, 255, 0.56)` --> `rgb(150, 150, 150)`

Same color, but no opacity issues.

**Screenshots**

Current:
<img width="204" alt="Screen Shot 2021-09-06 at 5 12 54 PM" src="https://user-images.githubusercontent.com/1312251/132263014-d91d867b-39ee-43ba-a57b-aad2cf94d12f.png">

Expected:
<img width="183" alt="Screen Shot 2021-09-06 at 5 20 45 PM" src="https://user-images.githubusercontent.com/1312251/132263314-dae7c966-6b32-45df-b18d-d8b0d544d998.png">

**Additional context**
I'm happy to spend a few minutes updating the colors and throwing up a PR, but I figured you'd be able to do something better than me just using a color picker and eyeballing it.

kdkasad:
I think it would be a good idea to use solid gray instead of semi-transparent white/black for the foreground colors.

PR #560, opened 2021-09-09
Rgba to rgb

adityatelange:

Because the theme's colors are set using opacity instead of actual color values, they interfere with elements other than text. For example, dark elements styled by entry-content use the --secondary color of rgba(255, 255, 255, 0.56). The text reads fine as a grey color against the black background, but if you add an emoji into your content it will also take on the 0.56 opacity value and look like it's fading away.

This PR includes: Same color, but no opacity issues.

The opacity of `go to top` button also changes here

[BUG] Emojis affected by theme opacity values Fixes: #557

## PR Checklist

### DIFF
--- a/assets/css/common/post-single.css
+++ b/assets/css/common/post-single.css
@@ -210,7 +210,7 @@
     display: block;
     margin: auto 0;
     padding: 10px;
-    color: rgba(255, 255, 255, 0.8);
+    color: rgb(213, 213, 214);
     background: 0 0;
     border-radius: 0;
     overflow-x: auto;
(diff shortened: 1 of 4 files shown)


adityatelange:
@aormsby @kdkasad Please check whether the colors look the same of different. Also let me know if there are any other visual differences

aormsby:
@adityatelange Looks pretty solid to me 👍

kdkasad (approved):
LGTM!

aormsby:
Just bumping this. I'd happily help deal with the conflicts, but I don't have write permission. 😅

PR #561, opened 2021-09-11
content can write formative code and change line

drizzl4:
before this update, you can't write a line and press 'enter' in content part  

so, you also can't write code with markdown grammar in content


adityatelange (approved):
Thanks :+1:

PR #614, opened 2021-10-30
Added missing comma in json

aheld:
When running a basic `hugo` build the build fails with a JSON parsing error
```
☁  aaronheld-blog [papermod] ⚡  hugo
Start building sites … 
hugo v0.84.0+extended linux/amd64 BuildDate=unknown
ERROR 2021/10/30 10:46:04 JSON parse error: expected comma character or an array or object ending on line 111 and column 40
   12:     {
           ^
```

also a stackoverflow question: https://stackoverflow.com/questions/68088702/json-parse-error-on-running-hugo-command

**The change**
this PR just adds a comma in a rendered JSON block for `<script type="application/ld+json">` in the `schema-json.html`template.
That item is an array, so it works with zero or 1 listItem, but with more then 1 you need a comma for valid JSON

## PR Checklist

### DIFF
--- a/layouts/partials/templates/schema_json.html
+++ b/layouts/partials/templates/schema_json.html
@@ -40,7 +40,7 @@
       "position": {{ add 1 $index  }},
       "name": {{ $bc_pg.Name }},
       "item": {{ $bc_pg.Permalink | safeHTML }}
-    }
+    },
     {{- end }}
 
   {{- end }}


danielfdickinson:
Are you sure this isn't because you are missing the trailing slash ('/') in baseURL in ``config.toml`` -- I made the same mistake and #559 discusses why the PR likely won't be accepted.

aheld:
> Are you sure this isn't because you are missing the trailing slash ('/') in baseURL in `config.toml` -- I made the same mistake and #559 discusses why the PR likely won't be accepted.

I think you are correct - I'll close this and research more thank you!

this works inconsistently between `hugo` and `hugo server`, so the baseurl makes sense

aheld:
confirmed for me that a `baseurl` in `config.toml` set to "/" causes the error

a blank baseurl or an actual url work fine

Error
`baseurl: "/"`

working
`baseurl: ""`
or
`baseurl: "http://localhost/"`

PR #714, opened 2022-01-05
Fix share URL

grem11n:

Fixes URL for share buttons. Currently, if you try to use share buttons in the bottom of the page, you will get a wron link. For example, on Twitter:

Screenshot 2022-01-05 at 23 40 34

I believe, this is caused by an error in obtaining the page URL. [This answer](https://discourse.gohugo.io/t/how-to-get-the-url-of-the-current-page-with-the-rel-or-relref-function/29113/4) on the Hugo forum has some more details.

Quick test with the local server:

Screenshot 2022-01-05 at 23 43 23

I couldn't find a relevant issue, but I can create one if required.

## PR Checklist

### DIFF
--- a/layouts/partials/share_icons.html
+++ b/layouts/partials/share_icons.html
@@ -1,4 +1,4 @@
-{{- $pageurl := .Permalink }}
+{{- $pageurl := .Page.Permalink }}
 {{- $title := .Title }}
 
 {{- $.Scratch.Set "tags" ""}}


layouts/partials/share_icons.html
```diff
@@ -1,4 +1,4 @@
-{{- $pageurl := .Permalink }}
+{{- $pageurl := .Page.Permalink }}
```
adityatelange, line 1: `.Page` is depreciated and has been removed in Hugo `v0.92.0` https://github.com/gohugoio/hugo/releases/tag/v0.92.0

grem11n:
Thank you for heads up, @adityatelange ! I didn't know that `.Page` is going to be removed.
Could you advise, how the issue with share buttons can be fixed then? I've upgraded Hugo to the latest version (`v0.92.0`), but didn't resolve the original issue.

adityatelange:
> Thank you for heads up, @adityatelange ! I didn't know that `.Page` is going to be removed. Could you advise, how the issue with share buttons can be fixed then? I've upgraded Hugo to the latest version (`v0.92.0`), but didn't resolve the original issue.

Looks like you don't have a baseURL set in the config. Pls try setting it like `baseURL: https://site.name/`

PR #734, opened 2022-01-14
Fall back post author to '.Site.Author.name' when it is not set.

thiagowfx:
'.Site.Author' is a recognized Hugo site variable, c.f. [1].

It's also already used elsewhere within PaperMod, namely in rss.xml.

By falling back to this variable users' configs can be simplified, being
effectively unnecessary to set the author twice, in .Site.Params.author.

[1]: c.f. https://gohugo.io/variables/site/#site-variables-list

c.f. commit message.

No.

N/A

## PR Checklist

- [N/A] This change adds/updates translations and I have used the [template present here](https://github.com/adityatelange/hugo-PaperMod/wiki/Translations#want-to-add-your-language-).
- [x] I have verified that the code works as described/as intended - verified in my hugo site
- [N/A] This change adds a Social Icon which has a permissive license to use it.

### DIFF
--- a/layouts/partials/author.html
+++ b/layouts/partials/author.html
@@ -1,5 +1,5 @@
-{{- if or .Params.author .Site.Params.author }}
-{{- $author := (.Params.author | default .Site.Params.author) }}
+{{- if or .Params.author .Site.Params.author .Site.Author.name }}
+{{- $author := (.Params.author | default .Site.Params.author | default .Site.Author.name) }}
 {{- $author_type := (printf "%T" $author) }}
 {{- if (or (eq $author_type "[]string") (eq $author_type "[]interface {}")) }}
 {{- (delimit $author ", " ) }}


adityatelange:
It is not added as a fallback because `.Site.Author` is not implemented inside Hugo itself.

adityatelange:
Take a look at https://github.com/adityatelange/hugo-PaperMod/commit/24e5d82ecb4f28bb7b22a3a8d28633b6a1519c48 for more information.

adityatelange:
Looks like this is legit https://discourse.gohugo.io/t/site-author-usage/31459/8

But it also supports many authors which which needs to be added accordingly

thiagowfx:
Fair enough. Supporting multiple authors would be a new pattern here so it warrants some discussion first.

Would a comma separated list be reasonable?

E.g. 'Foo Bar, Alpha Bravo'

adityatelange:
> Fair enough. Supporting multiple authors would be a new pattern here so it warrants some discussion first.
> 
> Would a comma separated list be reasonable?
> 
> E.g. 'Foo Bar, Alpha Bravo'

We already have comma separated list implemented, just needs to be fit Names from `.Site.Author` in it.

thiagowfx:
Do you mean to just replace `.Site.Author.name` with `.Site.Author`?

PR #740, opened 2022-01-16
Add option to add word count to metadata. Closes #735

### LINKED ISSUE #735: FR: Add option to add Word Count to metadata
In addition to `showReadingTime` ("5 min"), it would be handy to have native support for `showWordCount` or similar ("100 words").

Rationale: Reading time is subjective and relative. Word count is absolute and a more reliable and traditional proxy to estimate how long a post is. It's not perfect, and there are variations across languages, hence an option would leave it to users to decide whether it's relevant to their use cases.

## Proof-of-concept

- Commit: https://github.com/thiagowfx/thiagowfx.github.io/commit/cf023216d3f8cef8957f3a98b2a92d575a521583
- Live example: https://thiagowfx.github.io/2021/12/hello-world-from-hugo/

It basically comes down to adding the following snippet to `post_meta.html`:

```
{{- if (.Param "ShowWordCount") -}}
{{- $scratch.Add "meta" (slice (i18n "words" .WordCount | default (printf "%d words" .WordCount))) }}
{{- end }}
```

...plus adding translations for "words".

And the metadata looks something like:

```
December 11, 2021 · 1 min · 28 words · Thiago Perrotta
```

Happy to send a more polished PR than the aforementioned commit if the developers are willing to accept this contribution.

thiagowfx:
Reading time is subjective and relative. Word count is absolute and a
more reliable and traditional proxy to estimate how long a post is. It's
not perfect, and there are variations across languages, hence an option
would leave it to users to decide whether it's relevant to their use
cases.

Introduce a ShowWordCount option and use it in post_meta.html.

Hugo supports word counts natively, c.f.
https://gohugo.io/variables/page/#page-variables

c.f. commit description above

See https://github.com/adityatelange/hugo-PaperMod/issues/735

## PR Checklist

- [x] I have verified that the code works as described/as intended. --> tested on my blog with 'en' and 'pt', verified it works as expected
- [N/A] This change adds a Social Icon which has a permissive license to use it.

### DIFF
--- a/i18n/de.yaml
+++ b/i18n/de.yaml
@@ -9,6 +9,11 @@
     one: "1 min"
     other: "{{ .Count }} min"
 
+- id: words
+  translation:
+    one : "Wort"
+    other: "{{ .Count }} Wörter"
+
 - id: toc
   translation: "Inhaltsverzeichnis"
 
--- a/i18n/en.yaml
+++ b/i18n/en.yaml
@@ -9,6 +9,11 @@
     one : "1 min"
     other: "{{ .Count }} min"
 
+- id: words
+  translation:
+    one : "word"
+    other: "{{ .Count }} words"
+
 - id: toc
   translation: "Table of Contents"
 
--- a/i18n/es.yaml
+++ b/i18n/es.yaml
@@ -9,6 +9,11 @@
     one : "1 min"
     other: "{{ .Count }} min"
 
+- id: words
+  translation:
+    one : "palabra"
+    other: "{{ .Count }} palabras"
+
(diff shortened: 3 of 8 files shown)


thiagowfx:
Notes:

- I only added translations that I am confident to be correct. If you'd like I could add them all with google translate but I wouldn't be sure whether they are accurate.
- I couldn't find anywhere to document this option in. If there's a suitable to do so let me know, happy to send docs alongside this PR.

adityatelange (approved):
Thanks @thiagowfx!

PR #812, opened 2022-03-04
Fix multilingual tags not showing up in self-defined taxonomy and multilingual search issue

AidenKitamura:
I found that originally, only taxonomies with name "tags" would show up in posts. However this brings a problem: if one created a multilingual site with self-defined taxonomy, namely not using "tags" but the corresponding word in their own languages as the taxonomy term, "etiquetas" in Spanish, for example, then the tags will never show up in posts properly. This issue was not discussed before.

If one used "tags" as the taxonomy term, then the tag page's title will be "tags" instead of its translation. Imagine a page full of Spanish tags with an English title "Tags", while other entries remain Spanish. Weird, isn't it?

I tried to fix the problem by changing the {{.Title}} keyword in terms.html to {{i18n .Title}}, but it didn't fix the problem fully cause the head still uses .Title instead of its i18n version.

So in the end I changed the taxonomy layout. I made two changes:

1. Changed {{- if .Params.tags}} to {{- if isset .Params (i18n "tags" | default "tags")}} to check if the user has defined a different name for "tags" as their own taxonomy term.
2. Changed {{- range ($.GetTerms "tags") }} to {{- range ($.GetTerms (i18n "tags" | default "tags")) }} so that the tags could be ranged over properly.

These changes will allow user to use Taxonomy terms of "tags" in their own languages and at the same time, tags will appear on their posts correctly. For example, if a Spanish used "etiquetas" as taxonomy term for tag, the etiquetas will not show up in the posts at all before this change. However, after these changes, the etiquetas would show up properly. However, the user need to check the i18n translation files to find/set their taxonomy terms accordingly.

I tried to add corresponding translation in the i18n file except for central Kurdish and Bengali because I am unable to Google translate them.

## PR Checklist

### DIFF
--- a/assets/js/fastsearch.js
+++ b/assets/js/fastsearch.js
@@ -48,7 +48,7 @@ window.onload = function () {
             }
         }
     };
-    xhr.open('GET', "../index.json");
+    xhr.open('GET', jsonLangUrl); // Using Relative Path to Base URL instead of to current directory
     xhr.send();
 }
 
--- a/i18n/ar.yaml
+++ b/i18n/ar.yaml
@@ -20,3 +20,6 @@
   
 - id: home
   translation: "الصفحة الرئيسية"
+
+- id: tags
+  translation: "العلامات"
--- a/i18n/bg.yaml
+++ b/i18n/bg.yaml
@@ -14,3 +14,6 @@
 
 - id: translations
   translation: "Преводи"
+
+- id: tags
+  translation: "етикети"
--- a/i18n/ca.yaml
+++ b/i18n/ca.yaml
@@ -17,3 +17,6 @@
 
 - id: home
   translation: "Inici"
+
+- id: tags
+  translation: "Etiquetes"
--- a/i18n/da.yaml
+++ b/i18n/da.yaml
@@ -26,3 +26,6 @@
 
 - id: code_copied
   translation: "kopieret!"
+
+- id: tags
+  translation: "tags"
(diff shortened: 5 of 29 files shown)


AidenKitamura:
I changed _default/search.html and fastsearch.js to using relLangUrl function provided by hugo so that the search functions works on multilingual site even behind a CDN. I have already tested its functionality by deploying my own site on CloudFlare Pages (kitamura.moe) with CloudFlare CDN.

adityatelange (commented):
Hello @AidenKitamura, 

We are using `tags` just like another keyword we use to define a parameter.
https://github.com/adityatelange/hugo-PaperMod/blob/34127c9170ecf6fb2b274a0b2ade96088b551a47/content/posts/emoji-support.md?plain=1#L1-L9

For ex in the above lines we have `title`, `description`, `author` all of them as a variable to hold the value for front-matter.

The implementation you have here is expecting users to use a variable in their own laguage.

For ex. For Ukrainian it would be
```
title: A post in Ukrainian
теги: [sometag, someothertag]
```

This is not how it should be.

adityatelange:
Even if you still want that way you can use https://gohugo.io/content-management/taxonomies/#default-taxonomies

```yml
taxonomies:
  tag: теги
```

adityatelange:
This https://github.com/adityatelange/hugo-PaperMod/commit/4dc6b449562cc4f8046d22d1f844de2eaa556bba should help

PR #861, opened 2022-03-30
Modify the way paperMod displays project covers by putting them next to the titles

Axthauvin:
*First, I would like to say this is my first pull request, so sorry if I don't do it right :/*

## **What does this PR change? What problem does it solve?**

At the time I write this, PaperMod main page look like this :

### As we can see, the ***project cover*** takes the whole page size

So, I thought it would be more easy, and “clean” if projects cover looked like this :

Here, covers are displayed next to the title and the post content, which is more convenient
Furthermore, ***screen size is the same***, so that way we can save a lot of unnecessary scrolling…

## PR Checklist

## Acknowledgements
**This change was kind of proposed by **@nicopcat**
[As we can see here : https://github.com/adityatelange/hugo-PaperMod/discussions/691](https://github.com/adityatelange/hugo-PaperMod/discussions/691)**

*Thank you for reading my pull request, hoping to be helpful !*
🇫🇷


Axthauvin:
Ahah first pull request experiences i guess

adityatelange:
Hello @Axthauvin welcome to world of PRs :wink: 

Nice implementation you have here. I don't want to disappoint you but, we'd like to stick to the way it is implemented right now.
The size of cover image is also reflected over opengraph and twitter cards as well as Json schema for rich results over the web.
And these have a recommended image aspect ratio which is 4:3 in most of the cases. People should be using 1200x628 or somewhat similar size for cover images for better results.

Hope you understand. Thanks for contributing!

PR #924, opened 2022-05-23
mark nav item as active for sub-pages

makkes:

When navigating to e.g. a single blog page, the nav item this page is in is now marked as active so users know in which area of the site they are navigating at the moment.

No

## PR Checklist


makkes:
I had breadcrumbs deactivated and in that case it'd still be nice to show users where they are but I'm fine closing the PR.

PR #934, opened 2022-06-01
Allow Overriding of article in list.html

loyayz:

Motivation: I'd like to change the `<article>`  part of the `list.html`.
Problem: I must change list if want to change article,because the article  is currently defined in the list

No

## PR Checklist


adityatelange:
Sorry but,Seperation of templates into multiple files for personal convenience will make maintaining the theme complicated. Closing

PR #979, opened 2022-07-12
Ability to stick header to the top of the page

Wirone:

If enabled (using `stickyHeader` param), header with top menu will be always visible regardless of scrolling.

[Live Demo](https://blog.codito.dev/).

- Works on desktop and mobile
- Supports dark/light modes
- 100% backward compatible

No.

## PR Checklist

### DIFF
--- a/assets/css/common/header.css
+++ b/assets/css/common/header.css
@@ -1,3 +1,11 @@
+header.header-sticky {
+    position: sticky;
+    top: 0;
+    width: 100%;
+    background: var(--theme);
+    z-index: 1000;
+}
+
 .nav {
     display: flex;
     flex-wrap: wrap;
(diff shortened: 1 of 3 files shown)


ngntrgduc:
Thanks for the great feature~. But it will hide some content after clicking on the `Table of Contents`.

Here is my workaround: add this to the `assets/css/common/header.css`

```css
html {
    scroll-padding-top: calc(var(--header-height) * 2);
}
```

I hope this help.

Wirone:
@ngntrgduc Thanks for the great tip! I encountered this issue on my blog, but ignored it as I did not want to spend my time on it, however I've just applied your snippet and it really helps 🙂. Just added it to this PR.

I am not up-to-date with modern CSS tricks, so let me know if there's a way of applying this to `html` conditionally when `header` has `header-sticky` class included. Maybe there's no need to apply `header-sticky` also to `html` 😉.

PR #1008, opened 2022-08-16
Charts+graphs

divad1196:
I have read what comes before about not using CDN. I am wondering why, could you explain?
I have made some changes for my personal use and thought it could be useful for others, so I isolated my commits and made this PR. The CDN are only loaded if needed, and this is better than having a copy of the file.

I am open to discussion and I am ready to collaborate if it can help.

This PR allows:
* Use of mermaid diagrams (shortcode)
* Use of charts (shortcode)
    * This use Chart.js (cdn)
    * We can provide the whole parameters to define the exact diagram we want
    * There is a simplified version for barCharts. Colors are predefined but can be override

Not as far as I know

## PR Checklist

### DIFF
--- a/data/charts/colors.yml
+++ b/data/charts/colors.yml
@@ -1,0 +1,28 @@
+red: &red '#FF6384FF' 
+bgRed: &bgRed '#FF638433'
+blue: &blue '#36A2EBFF'
+bgBlue: &bgBlue '#36A2EB33'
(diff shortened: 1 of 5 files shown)


divad1196:
Closing this PR:
- Answer taking to long
- Don't want to contribute after seeing the kind of responses provided when they are some.

PR #1011, opened 2022-08-18
i18n: Improve date localization

RoneoOrg:
Hugo recently improved the way dates are displayed in languages other than english.
Let's implement this feature. It requires to bump to required Hugo version to 0.87.
Ref: https://gohugo.io/functions/dateformat/


adityatelange:
- #555 
Already deals with it and is included, thanks

PR #1012, opened 2022-08-19
Adding threema as social icon

pabloscloud:
It adds the theema svg from https://github.com/simple-icons/simple-icons to the social-icons

I don't think so


adityatelange (approved):

pabloscloud:
thanks :)

Issue #962, opened 2022-06-29
Content with Cover Images generate Cumulative Layout Shift


jaxn020:
Dear Adithya, kindly check on this.
Posts without any Cover Image ( Ex: https://adityatelange.github.io/hugo-PaperMod/posts/papermod/papermod-installation/) show ZERO CLS score (which is perfect!),
But as soon as an Image is added (as cover image), either TOC creates a shift, or when TOC is false, post-content creates a shift or sometimes both. The CLS score is >0.2 which hurts site metrics. Please throw some light.
By the way, thanks for this Wonderful Theme!

adityatelange:
Can you help me why fix this? Pls investigate why this is happening.

sunuazizrahayu:
> Can you help me why fix this? Pls investigate why this is happening.

it's happening because we're not defining image size on attribute.
it might helpful [https://gtmetrix.com/cumulative-layout-shift.html](https://gtmetrix.com/cumulative-layout-shift.html)

adityatelange:
> > Can you help me why fix this? Pls investigate why this is happening.
> 
> @adityatelange ,it's happening because we're not defining image size on attribute. it might helpful https://gtmetrix.com/cumulative-layout-shift.html

We are when using page bundles. 
https://github.com/adityatelange/hugo-PaperMod/blob/045c08496d61b1b3f9c79e69e7d3d243a526d8f3/layouts/partials/cover.html#L27

there's no wat to calc it if you're statically linking it instead of using page bundles.

PR #1014, opened 2022-08-21
feat(layouts): add iconWidth support for icon

Airscripts:

Adds support for a new field in config's params called iconWidth.  
With iconWidth set, Lighthouse will not complain about Cumulative Shift Layouts.  
More on topic [here](https://web.dev/cls/).  

No, it wasn't treated about icons specifically.  

## PR Checklist

### DIFF
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -43,41 +43,60 @@
     <nav class="nav">
         <div class="logo">
             {{- $label_text := (site.Params.label.text | default site.Title) }}
+            
             {{- if site.Title }}
-            <a href="{{ "" | absLangURL }}" accesskey="h" title="{{ $label_text }} (Alt + H)">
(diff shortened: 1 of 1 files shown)


adityatelange:
@Airscripts Height is enough for resizing the icon keeping the aspect ratio intact.
What we can do for cls thing here is print the width of image, which is already available in hugo context.
We can add the param for images where the width isn't calculable.

Airscripts:
@adityatelange I've removed the resize on width and tested it up.  
In the meantime I've even cleaned up the code a little bit, just to make it more readable and maintainable.  

About the #962 I can look at it for sure.  
I'll let you know what it needs to be done in order to work properly and eventually I'll open a dedicated PR.  

Thanks!

adityatelange (changes requested):
@Airscripts can you please revert to previous code style, will make it easier to track the changes

PR #1105, opened 2022-12-12
add title to the image tag in rss.xml

k-yone:
* The `<title>` tag is missing in the `<image>` tag in rss.xml
* An error message is returned when you try to register the rss.xml to Google Search Console.
* Here is the RSS2.0 specification. It says it's required.
  * https://cyber.harvard.edu/rss/rss.html#ltimagegtSubelementOfLtchannelgt

* I guess not?

## PR Checklist

### DIFF
--- a/layouts/_default/rss.xml
+++ b/layouts/_default/rss.xml
@@ -18,6 +18,7 @@
     <description>Recent content {{ if ne  .Title  site.Title }}{{ with .Title }}in {{.}} {{ end }}{{ end }}on {{ site.Title }}</description>
     {{- with site.Params.images }}
     <image>
+      <title>{{ site.Title }}</title>
       <url>{{ index . 0 | absURL }}</url>
       <link>{{ index . 0 | absURL }}</link>
     </image>


adityatelange (commented):
the title described here https://cyber.harvard.edu/rss/rss.html#ltimagegtSubelementOfLtchannelgt

says:
> <title> describes the image, it's used in the ALT attribute of the HTML  tag when the channel is rendered in HTML.

Is not the site's title

adityatelange:
Thing is we couldn't get the image's alt text which is inside the markdown with the `images` front matter variable.

k-yone:
I see. You're right.
And the problem is, this <image> element is used to describe the whole website so there aren't any `.md` files that the user can declare the alt text for this image in a front matter.

The document says this <image> element is meant to be used as a link to the site, so, personally, I think it's not completely wrong to use the site tile for the alt text, but it's nicer to add an option to the config to let the user to set an alt text.

k-yone:
By the way, this <link> has to be a link to the website, not the image's url, right? I didn't notice.
https://github.com/adityatelange/hugo-PaperMod/pull/1105/files#diff-3fb10c6a0e1801f8c8a23a6858a8710151ae674fd98589294786ecb97e8220e2R23

adityatelange:
> By the way, this `<link>` has to be a link to the website, not the image's url, right? I didn't notice. https://github.com/adityatelange/hugo-PaperMod/pull/1105/files#diff-3fb10c6a0e1801f8c8a23a6858a8710151ae674fd98589294786ecb97e8220e2R23

Nope, the link is for each image provided by user.

adityatelange:
@k-yone I was wrong. It is used to describe the whole website. We can merge.

adityatelange:
For RSS feeds present inside a section we'll have to add something, but that is a different topic to be discussed.

adityatelange (approved):

PR #1123, opened 2023-01-06
Modify breadcrumbs logic

dsegonds:

The breadcrumbs generation algorithm breaks when the base URL is "/".

I do not believe there was a related issue or discussion before.

## PR Checklist

## Explanation

The breadcrumbs generation algorithm's first part is removing the base URL from the parent's permalink URL.

In the existing code, this is achieved by replacing all instances of the base URL in the parent's permalink URL by 
an empty string.

This algorithm fails to produce breadcrumbs when the base URL is "/", for example.

This is erroneous as only the first instance of the base URL, starting at the beginning of the parent's permalink URL, should be replaced by an empty string.

This is why I replaced the call to "replace" with a call to "replaceRE".


adityatelange:
Fixed by #1368

PR #1159, opened 2023-02-13
feat: new language switch behaviour

### LINKED ISSUE #1065: Language selector to forward to translated page
Current state: This theme only supports the header language selector to forward to the start page. Links forwarding to the translation of the page you are currently on is only available with blog entries.

Expected behaviour: Either the default or a variable enables the header language selector to forward directly to the translation of a particular page; if no translation exists, the language selector shall forward to the start page. This behaviour should work with all types of web pages.

Advantages: Simplification of navigation. Improved multilingual UX.

Disadvantages: Either changes to default behaviour or one additional variable to keep in mind and document.

Generally speaking, thank you to the maintainer and all contributors for this great Hugo theme!

HarisDotParis:

Currently, the language switch _always_ links to each language's main page and not any possible translation of the page. This PR changes the way the language switch works significantly. The language switch still lists all languages available according to their "weight", but if there is a translation, then that translation will be linked, otherwise that language's main page as before. As this replaces the need for the translation list in posts, translation lists in posts have been made optional with a new boolean parameter "showTranslationListInPosts". I have tested the changes extensively and with different parameters and language settings.

Closes #1065 

## PR Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@
 -   Social Icons (home-info and profile-mode)
 -   Social-Media Share buttons on posts.
 -   Menu location indicator.
--   Multilingual support. (with language selector)
+-   Multilingual support (with improved language selector)
 -   Taxonomies
 -   Cover image for each post (with Responsive image support).
 -   Light/Dark theme (automatic theme switch a/c to browser theme and theme-switch button).
--- a/layouts/_default/search.html
+++ b/layouts/_default/search.html
@@ -13,10 +13,10 @@
         {{ .Description }}
     </div>
     {{- end }}
-    {{- if not (.Param "hideMeta") }}
-    <div class="post-meta">
-        {{- partial "translation_list.html" . -}}
-    </div>
+    {{- if and (.Param "showTranslationListInPosts") (not (.Param "hideMeta")) }}
+        <div class="post-meta">
+            {{- partial "translation_list.html" . -}}
+        </div>
     {{- end }}
 </header>
 
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -15,7 +15,9 @@
     {{- if not (.Param "hideMeta") }}
     <div class="post-meta">
       {{- partial "post_meta.html" . -}}
-      {{- partial "translation_list.html" . -}}
+      {{- if .Param "showTranslationListInPosts" }}
+        {{- partial "translation_list.html" . -}}
+      {{- end }}
       {{- partial "edit_post.html" . -}}
       {{- partial "post_canonical.html" . -}}
     </div>
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -96,27 +96,42 @@
                 </button>
                 {{- end }}
 
-                {{- $lang := .Lang}}
-                {{- $separator := or $label_text (not site.Params.disableThemeToggle)}}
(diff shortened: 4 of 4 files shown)


adityatelange:
Hi @HarisDotParis there is a reason that the language switch always links to each language's main page and not any possible translation of the page.
And it is: regularity. The website should have regularity in the way it behaves upon a certain action.

For the case where the particular page has a translated page listed, we have the translations, mentioned right under the title of the page
image
So that if the user wants to switch to their preferred language version,they can happily move there.

HarisDotParis:
I am sorry, but I consider this deviating from the regular behaviour of language switches. Most multilingual websites forward to the translation if possible, and only to the main page if no translation is available. This also makes sense, because most users switching languages want to read the same content in either their own or the original language.

Regularity only for the sake of it doesn't make sense. Always forwarding to the main page may create regularity, but it makes a simple feature more complex, and this makes the UX worse.

algo-se:
I think the site would have greater regularity if there was only one switch (in the nav bar) and not two (nav bar + posts meta), but then the switch in nav bar should translate the page and not go to home. That said, it's not a big deal, just my 2 cents. 

Awesome theme btw! 😄

cubxxw (approved):

jakoblover:
Any updates on this? Would love this feature. Having two translation buttons is very confusing.

vwegert:
I completely agree with @HarisDotParis and @algo-se - in fact, I came here to report the current behaviour as a bug. For some reason, the translation links below the header are not shown in my "playground" site, but even if they were, the user should not be thrown back to the index page just by selecting a different language. I can't see I've ever encountered a page that behaves this way.

vwegert:
The PR might need to be updated - I'm getting the error message 
```
ERROR deprecated: .Site.IsMultiLingual was deprecated in Hugo v0.124.0 and subsequently removed. Use hugo.IsMultilingual instead.
```
when applying the changes locally.

Drumclem:
This is EXACTLY what I was looking for. I love the theme, but that language toggler behaviour bugged me to no end :)

Also: I had an error come up because you're supposed to write hugo.IsMultilingual and not hugo.IsMultiLingual

Hope that helps!

Also also: the flag still shows up in sections, but I assume it won't be too complicated to change. Thanks @HarisDotParis !

jakoblover:
Why was this closed? The way the language selector currently works in papermod is useless.

adityatelange:
> Why was this closed? The way the language selector currently works in papermod is useless.

Don't use it if your find it useless. Do something useful then.

miallo:
I just keep rebasing this minimal patch:
```diff
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -58,7 +58,11 @@
                 {{- if (not site.Params.disableLangToggle) }}
                     {{- $lang := .Lang}}
                     {{- $separator := or $label_text (not site.Params.disableThemeToggle)}}
-                    {{- with site.Home.Translations }}
+                    {{- $translations := .Translations }}
+                    {{- if not $translations }}
+                        {{- $translations = site.Home.Translations }}
+                    {{- end }}
+                    {{- with $translations }}
                     <ul class="lang-switch">
                         {{- if $separator }}<li>|</li>{{ end }}
                         {{- range . -}}
```
Works fine for me

PR #1169, opened 2023-03-08
Escape ampersand in HTML

bkrl:

This PR fixes an unescaped ampersand in the HTML templates. The page renders fine without this but ampersands should be escaped in HTML.

No

## PR Checklist


bkrl:
I just found out that HTML5 allows unescaped ampersands in some cases and Hugo's minifier actually unescapes this ampersand. I guess this ampersand doesn't need to be escaped then.

PR #1181, opened 2023-03-28
change <del> element CSS styling to use line-through

dubbl:

I used the markdown `~~strikethrough~~` markup in a blog post, and noticed that it was rendering weirdly in Firefox 111:

Image showing a very thin and broken line through text

In Chrome it rendered correctly:

Image showing normal strike-through text

It turns out, that hugo converts `~~` to the [HTML `del` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/del). This element was apparently not supported in Chrome until recently, which is why I guess this fix was introduced in hugo-Paper back in the day. hugo-Paper removed the original linear-gradient solution [in v6 released in July 2023](https://github.com/nanxiaobei/hugo-paper/commit/39f6f4e9b2f6d6beed74cc46325bc0157a40a2b1).

According to caniuse.com [only around 69% of mobile user and 22% desktop users use a browser that supports this element](https://caniuse.com/mdn-html_elements_del).
I assume that this will change rapidly with the Chrome support (Version 111, released March 7th 2023).
But I would still propose that we style this element manually, as it can be quite important to indicate that a phrase is no longer up to date/valid.

Therefore this change uses `text-decoration: line-through`  to archive the same effect across browsers.

After fix in Firefox 111:
Correctly strike through text

After fix in Chrome 111:
Correctly strike through text

I couldn't find anything.

## PR Checklist

### DIFF
--- a/assets/css/common/post-single.css
+++ b/assets/css/common/post-single.css
@@ -83,8 +83,7 @@
 }
 
 .post-content del {
-    text-decoration: none;
-    background: linear-gradient(to right, var(--primary) 100%, transparent 0) 0 50%/1px 1px repeat-x;
+    text-decoration: line-through;
 }
 
 .post-content dl,


adityatelange:
Thanks @dubbl

PR #1156, opened 2023-02-11
Add RSS icon to archive pages

cljoly:
**What does this PR change? What problem does it solve?**

If the setting “ShowRssButtonInSectionTermList” is true, I think it
makes sense to also show the RSS icon by the title in archive pages, as
we do for other article lists.

Tested on [my website](https://cj.rs/blog/):
image

This reuses the code of `layouts/_default/list.html`.

**Was the change discussed in an issue or in the Discussions before?**

Not that I’m aware.

## PR Checklist

- [X] I have enabled [maintainer edits for this PR](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [X] I have verified that the code works as described/as intended.
- [X] This change **does not** include any CDN resources/links.
- [X] This change **does not** include any unrelated scripts such as bash and python scripts.

PR #1183, opened 2023-04-01
Revert "Add RSS icon to archive pages"

adityatelange:
Reverts adityatelange/hugo-PaperMod#1156


totoroot:
@adityatelange Can you please explain why this was reverted?

adityatelange:
> @adityatelange Can you please explain why this was reverted?

Patch does not work as it should, incompatible with Papermod. The author has modified their theme to use the archive page on the home page.

PR #1192, opened 2023-04-18
Fix RSS link for paginated list pages

bkrl:

The RSS link in list pages was previously hard-coded to `index.xml`, which doesn't work on paginated list pages because it might go to something like `/section/pages/2/index.xml` instead of `/section/index.xml`. I made it use `(.OutputFormats.Get "rss").RelPermalink` instead following the [example in Hugo's docs](https://gohugo.io/templates/rss/#reference-your-rss-feed-in-head).

## PR Checklist


adityatelange (approved):
Perfect!

PR #1201, opened 2023-04-28
feat: Option to hide logo label text

percula:
This PR adds the option to hide the logo label text. This is useful when the logo image or SVG already contains the site's name.

I structured the change so that it's backwards-compatible; the logo text will only be hidden if `params.label.showText` is set to false.

image

Not that I'm aware of.

## PR Checklist

### DIFF
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -69,7 +69,9 @@
                 {{- else if hasPrefix site.Params.label.iconSVG "<svg" }}
                     {{ site.Params.label.iconSVG | safeHTML }}
                 {{- end -}}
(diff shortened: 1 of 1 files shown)


percula:
Well, I didn't think about doing that. That would work, but it would affect the spacing between the logo and the light/dark toggle.

adityatelange:
@percula I'd request you to create issues before PRs, many times, efforts may not yield the intended results and PR might not get merged.

adityatelange:
> 

good

adityatelange:
Many a times not reviewing a PR is better than reviewing, without expecting anything in return.

PR #1202, opened 2023-04-28
Add menu to footer

percula:
This PR will show a menu within the footer, if set within the site's params.

```yaml
    menu:
      main:
       ...
      footer:
        - name: Footer menu option
          url: "menuUrl"
          weight: 1
```

For now, this uses the same exact code as the header menu, just copy/pasted into the footer. I did have to modify the CSS a bit to remove the underline from the menu links, while keeping it on the other links in the footer. In the future, we may want to center-align the menu, but I feel this is good for now.

Example:
image

Not that I know of.

## PR Checklist

### DIFF
--- a/assets/css/common/footer.css
+++ b/assets/css/common/footer.css
@@ -58,3 +58,11 @@
 #theme-toggle:focus {
     outline: 0;
 }
+
+.footer nav a {
+    border-bottom: none;
+}
+
+.footer nav a:hover {
+    border-bottom: none;
+}
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -1,5 +1,34 @@
 {{- if not (.Param "hideFooter") }}
 <footer class="footer">
+    <nav class="nav">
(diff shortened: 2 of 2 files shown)


adityatelange:
I don't think this is useful, goes out of the design structure.

percula:
Really? Seems pretty useful to me. Pretty much every site has a menu in the footer. Plus, this feature is completely optional and won't change the appearance unless someone wants to add a footer menu.

adityatelange:
> Really? Seems pretty useful to me. Pretty much every site has a menu in the footer. Plus, this feature is completely optional and won't change the appearance unless someone wants to add a footer menu.

Menu in footer might be useful but its not just related to adding menu. Try considering other factors like:
- displaying it in mobile view
- making menus accessible in mobile view by not confusing user which one is meant for navigation and which is just for suporting links
- We support multilingual blog, how would it look with langs which are right aligned
- Footer menu overlapping the go to top button
- maybe more

PR #1238, opened 2023-06-16
Use the site author in copyright footer if available

morxa:

If no custom copyright statement is provided but the site has an `author` site parameter, then use the `author` instead of the page title in the copyright statement.

Having an author site parameter is fairly standard: https://gohugo.io/variables/site/#the-siteparams-variable

no

## PR Checklist

### DIFF
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -3,7 +3,7 @@
     {{- if site.Copyright }}
     <span>{{ site.Copyright | markdownify }}</span>
     {{- else }}
-    <span>&copy; {{ now.Year }} <a href="{{ "" | absLangURL }}">{{ site.Title }}</a></span>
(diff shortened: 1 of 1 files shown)


adityatelange (commented):
You can customize the whole Copyright string by setting `site.Copyright`
IMO we don't need of defaulting to author name.

morxa:
Yes I'm aware that you can override the whole copyright string, but customizing the whole copyright string is unnecessary work if you want to use the standard copyright string, just with an author.

Note that even the example site has "© 2023 [PaperMod](https://adityatelange.github.io/hugo-PaperMod)", which is questionable, because from my understanding, "PaperMod" cannot hold a copyright. Using the author seems to be a more reasonable default and it does not break anything as the title is still used if the author is not set.

I'd appreciate it if you could reconsider this PR!

adityatelange:
Commit https://github.com/adityatelange/hugo-PaperMod/commit/28b86cb5a602f4a64e90dce035e30ebe9fad1d6a might help to solve this challenge.

PR #1250, opened 2023-06-30
feat: allow hiding author in post_meta

0x2E:

For a person's site, in most cases the author is the person, and displaying it on every post and list may be redundant.

Setting the `params.author` in the configuration to empty will achieve the hidden effect, but it will also empty the value of `<meta name="author">`.

So add `ShowAuthor`.

none

## PR Checklist


adityatelange (commented):
It would be better if we keep showing author by default.
To allow hiding of author (is user wants), instead of `ShowAuthor`, `hideAuthor` would be they key.

0x2E:
It has been renamed to `hideAuthor`.

adityatelange (approved):

PR #1287, opened 2023-08-14
Localize menu items with translation tables

paradoja:

In the [Hugo documentation](https://gohugo.io/content-management/multilingual/#use-translation-tables) it explains using translation tables as a possible way of localizing menu entries.

This basically uses the menu identifier as a language key for the name of the entry, and is present in the [example menu template](https://gohugo.io/templates/menu-templates/#example).

This change basically gives the same functionality to PaperMod.

Not really. I searched for it and was going to ask. But the change is so minuscule that I thought opening the PR would be easier (and if there's a discussion it can happen anyway here).
## PR Checklist

### DIFF
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -130,7 +130,7 @@
                 {{- cond $is_search (" accesskey=/" | safeHTMLAttr) ("" | safeHTMLAttr ) }}>
                     <span {{- if eq $menu_item_url $page_url }} class="active" {{- end }}>
                         {{- .Pre }}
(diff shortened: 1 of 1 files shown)


adityatelange:
Adding menu translations at translation tables and rest of config in config.yml may have very less use cases.
Instead we are adding them to config  itself like this:

https://github.com/adityatelange/hugo-PaperMod/blob/4bdc993b98cc82d1c59efca0774f85811b0bf213/config.yml#L22-L67

PR #1290, opened 2023-08-15
Edit farsi translation and add 'words'

hialisabet:
I suggest more comfortable translation for 'prev_page' and 'next_page'. Also, I added farsi translation for missing 'words'.

There are more comfortable farsi translations for 'prev_page' and 'next_page'. also the translation of 'words' was missing in farsi. because of different directions, It ruins the style of the text!

No, I just noticed and decided to solve it.

## PR Checklist

### DIFF
--- a/i18n/fa.yaml
+++ b/i18n/fa.yaml
@@ -1,19 +1,24 @@
 - id: prev_page
-  translation: "صفحه قبلی"
+  translation: "صفحه قبل"
 
 - id: next_page
-  translation: "صفحه بعدی"
+  translation: "صفحه بعد"
 
 - id: read_time
   translation:
-    one: "۱ دقیقه"
+    one : "1 دقیقه"
     other: "{{ .Count }} دقیقه"
+
+- id: words
+  translation:
+    one : "کلمه"
+    other: "{{ .Count }} کلمه"
 
 - id: toc
   translation: "فهرست مطالب"
 
 - id: translations
-  translation: "ترجمه ها"
+  translation: "ترجمه‌ها"
 
 - id: home
   translation: "خانه"


adityatelange (commented):
Hi @pikhosh @mehdy, can you please help reviewing? Thanks.

i18n/fa.yaml
```diff
@@ -1,19 +1,24 @@
...
+  translation: "صفحه قبل"
 
 - id: next_page
-  translation: "صفحه بعدی"
+  translation: "صفحه بعد"
 
 - id: read_time
   translation:
-    one: "۱ دقیقه"
+    one : "1 دقیقه"
```
mehdy, line 9: Why would you want to change this? I think it's better to keep using the Persian numbers.

i18n/fa.yaml
```diff
@@ -1,19 +1,24 @@
...
     other: "{{ .Count }} دقیقه"
 
+- id: words
+  translation:
+    one : "کلمه"
```
mehdy, line 14: Shouldn't this be "۱ کلمه" or "یک کلمه"?

i18n/fa.yaml
```diff
@@ -1,19 +1,24 @@
...
+    other: "{{ .Count }} کلمه"
+
 - id: toc
   translation: "فهرست مطالب"
 
 - id: translations
-  translation: "ترجمه ها"
+  translation: "ترجمه‌ها"
```
mehdy, line 21: I think the correct form would be using half-space. Ref: https://utype.ir/blog/%D9%86%DB%8C%D9%85-%D9%81%D8%A7%D8%B5%D9%84%D9%87-%D9%81%D8%A7%D8%B1%D8%B3%DB%8C/#:~:text=%DB%B5%2D%20%D8%AF%D8%B1%20%D8%B2%D9%85%D8%A7%D9%86,%DA%A9%D8%AA%D8%A7%D8%A8%E2%80%8C%D9%87%D8%A7%D8%8C%20%D8%AE%D8%A7%D9%86%D9%87%E2%80%8C%D9%87%D8%A7

i18n/fa.yaml
mehdy: If you like, you could use "واژه" as well.

mehdy (changes requested):

i18n/fa.yaml
mehdy: @hialisabet Okay, I didn't notice it. But then I think it'd make sense to change the plural forms to show the Persian numbers.

i18n/fa.yaml
mehdy: I don't know if you've changed the code or if I was confused before. Anyway, you're doing the right thing now.

PR #1324, opened 2023-10-28
 Add logic to use .Description instead of .Summary

masonwilde:
Adds a check for a useDescription param which will check for a description field in a post and use it instead of the .Summary when producing list.html. If a description is unavailable and hideSummary is not true, the summary will show as normal.

Changes the logic in using the .Summary when creating a list of posts.

Allows users to specify a useDescription param to check for and use a short description instead of the summary in the entry-content div.

My own discussion https://github.com/adityatelange/hugo-PaperMod/discussions/1323

Somewhat related to https://github.com/adityatelange/hugo-PaperMod/discussions/148

## PR Checklist

### DIFF
--- a/layouts/_default/list.html
+++ b/layouts/_default/list.html
@@ -72,7 +72,11 @@
       {{- if .Draft }}<sup><span class="entry-isdraft">&nbsp;&nbsp;[draft]</span></sup>{{- end }}
     </h2>
   </header>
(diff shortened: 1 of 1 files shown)


adityatelange:
We can use `description` and `summary` both at the same time with same or different text in it. It seems unnessary to have more options to do just this.

PR #1333, opened 2023-11-15
Add ability to load more partials in header

iliajie:
Hello,

**What problem does it solve?**

This PR adds ability to load custom partials in different parts of `header.html`, so there will be no need in changing upstream (PaperMod) `header.html` file.

## PR Checklist

- [X] This change updates the overridden internal templates from HUGO's repository.

### DIFF
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -40,6 +40,9 @@
 {{- end }}
 
 <header class="header">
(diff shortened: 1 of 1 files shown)


adityatelange (commented):
This way each section would need a extended partial for user customization.
Sorry but no, this isn't practical.

PR #1354, opened 2023-12-10
add if condition to check series value

zwbgood6:


adityatelange:
This is part of hugo's templates will be fixed with 
- #1328

adityatelange:
fixed by #1328

Issue #296, opened 2021-03-07
KaTex & Mermaid Support


Sniperhid:
Firstly thanks for this nice theme! I am going to use it for a new project.

One thing I am interested in having is support for KaTeX & Mermaid using shortcodes. This would involve including some javascript libraries to do it (e.g. KaTex & Mermaid JS).

I am happy to do the work to make this happen and submit a Pull Request soon - But before I do I am wondering if this would of interest/in scope of this repository?

kdkasad:
This has been discussed before, and the short answer is no. For the long answer, read @adityatelange's comment in #132 ([direct link](https://github.com/adityatelange/hugo-PaperMod/pull/132#issuecomment-745085672)). The goal of PaperMod is to be simple and have 0 outside dependencies. This rules out inclusion of KaTeX and other external libraries.

Here's how to enable KaTeX in a shortcode: bee-san@c5c9ffc3006f5fb499b2a8c66b94bdd3800433eb. There's also documentation of this on the [example site](https://adityatelange.github.io/hugo-PaperMod/posts/math-typesetting/).

Sniperhid:
Thanks @kdkasad for all the information, that's helpful!

I'll close this issue as it is clear it is out of scope.

PR #1356, opened 2023-12-12
add extend content

wirano:

Introduces an extent content hook, which is useful when adding custom codes below `.Content`.

e.g., adding mermaid diagrams, as [hugo doc](https://gohugo.io/content-management/diagrams/#mermaid-diagrams) described.

Now, we just need to create `/layouts/partials/extend_content.html` and add a snippet to it. There is no need to edit the theme that makes the submodule dirty.

Compared to adding a snippet to `extend_footer.html`, this hook only affects post pages that have mermaid code blocks, which may speed up loading.

P.S., I don't know what the name of this hook should be. Maybe there is a better choice.

#763 #850 #296 

## PR Checklist


Z10N0110:
Any update on this?

PR #1370, opened 2023-12-29
add cover image to rss

vindex10:

Related issue: #891

Include cover image into rss feed. Example in the screenshot:

image

This PR also restores the removed xmlns: ` xmlns:content="http://purl.org/rss/1.0/modules/content/"`
It was removed by https://github.com/adityatelange/hugo-PaperMod/commit/00488d01463fc8c8db9fbda3f6060a0b5b373464 , which broke the use of the `content:encoding` tag.

I chose to keep it as a single change, since the xmlns was an accidental regression, and this PR itself is small.

## PR Checklist

- [ ] (restored the CDN link which was used before) This change **does not** include any CDN resources/links.


vindex10:
@adityatelange now it is cleaner, after https://github.com/adityatelange/hugo-PaperMod/pull/1411 and I rebased to resolve the conflicts.

PR #1380, opened 2024-01-02
tags in rss displayed

vindex10:
Display tags in RSS feed.  Follows rss item cateogory spec: https://www.rssboard.org/rss-profile#element-channel-item-category

By default All site taxonomies are used. One however can choose taxonomies to include into RSS manually:

hugo.toml
```
params:
  TaxonomiesInRSS: ["tags", "categories"]    
```

## PR Checklist


thijsdezoete:
NICE!
I would propose to also add this to the json output in order for search to work too

Issue #30, opened 2020-10-08
[Discussion]Meta Tags for better Social Sharing of posts.


aghdom:
How would you feel about adding `meta` tags for various social networks to the `head.html` partial to make the shared posts better looking? 
The theme already supports easy sharing using the buttons, it would be nice to have the posts also look good on social media.

It should be a relatively easy addtion. (Here's an example of the tags for Twitter and Facebook): 
```html
<!-- Open Graph / Facebook -->
<meta property="og:type" content="website">
<meta property="og:url" content="https://example.com/">
<meta property="og:title" content="Post Title">
<meta property="og:description" content="Post Description">
<meta property="og:image" content="<cover-image-url>">

<!-- Twitter -->
<meta property="twitter:card" content="summary_large_image">
<meta property="twitter:url" content="https://example.com/">
<meta property="twitter:title" content="Post Title">
<meta property="twitter:description" content="Post Description">
<meta property="twitter:image" content="<cove-image-url>">

```

adityatelange:
Hello @aghdom,
We already have implemented og, twitter cards, one need to set `env` to production as mentioned in code below.
https://github.com/adityatelange/hugo-PaperMod/blob/2c7cffb617dd925d1e69432a34f6987314c2b8c1/layouts/partials/head.html#L50-L55

However the cover image linking is already under discussion https://github.com/adityatelange/hugo-PaperMod/issues/13#issuecomment-705313991

The linked pull-request https://github.com/adityatelange/hugo-PaperMod/pull/31 will add _duplicate_ code to the head section, we need more better implementation by overriding the hugo's internal template for both og as well as twitter-cards.

adityatelange:
I have added stuff for this change =>  https://github.com/adityatelange/hugo-PaperMod/compare/6c4e094c91ce%5E...306889f68bcb
But I am unsure if this needs to be merged :thinking: 

Hugo has a way to include the image in `images` 

ref : https://github.com/gohugoio/hugo/blob/e9a7ebaf67a63ffe5e64c3b3aaefe66feb7f1868/tpl/tplimpl/embedded/templates/opengraph.html#L5-L16

It can be used in social meta just by adding `cover` or `thumbnail` in its name and it will show up :tada: 

---

> this won't work as hugo's (opengraph and twitter)'s template doesn't support relative urls, and I cannot fix that

Ah damn! that's disappointing. Looks like there are some fixes scattered on the internet. I'll try these

[...]

Issue #13, opened 2020-10-03
Relative path for cover image


singhkays:
From the installation notes I see that the cover image supports absolute path. I'm trying to migrate from another theme where the following relative path for cover image works

`cover = "media/small-cover-spotify.png"`

which **resolves to** `http://localhost:1313/media/small-cover-spotify.png` 
**instead of** `http://localhost:1313/blog/how-spotify-is-killing-the-open-podcast-ecosystem/media/small-cover-spotify.png`

How can I get this working?

adityatelange:
Hey @singhkays, 
I have pushed a commit for this issue
https://github.com/adityatelange/hugo-PaperMod/commit/d8199140e5940f02960196d1220c9a0ce8a23de1

Please let me know if that works fine for you ...

singhkays:
@adityatelange works great now! I'm very excited to see you're making a lot of progress over the last few days on this theme 👏 I'll be making this the main theme on my blog https://www.singhkays.com/. I'll close this issue but post a couple of ideas as feature suggestions here on Github.

adityatelange:
> @adityatelange works great now! I'm very excited to see you're making a lot of progress over the last few days on this theme 👏 I'll be making this the main theme on my blog https://www.singhkays.com/. I'll close this issue but post a couple of ideas as feature suggestions here on Github. 

Thank you for your kind words :)
suggestions are welcome.

singhkays:
@adityatelange Looks like this is still a problem. Here's a live example - https://www.singhkays.com/blog/av1-2019-year-in-review/. The path should be `https://www.singhkays.com/blog/av1-2019-year-in-review/media/av1-2019-year-in-review.png` but instead is the below

Interestingly the actual path to the cover image that is displayed on that page is (_**Note the /./ in the path**_)
`https://www.singhkays.com/blog/av1-2019-year-in-review/./media/av1-2019-year-in-review.png`

```html
<meta property="og:url" content="https://www.singhkays.com/blog/av1-2019-year-in-review/" />
<meta property="og:image" content="https://www.singhkays.com/media/av1-2019-year-in-review.png" />
<meta name="twitter:card" content="summary_large_image"/>
<meta name="twitter:image" content="https://www.singhkays.com/media/av1-2019-year-in-review.png"/>
```

Here's how the cover and images paths are specified in the front-matter

```yaml
cover: 
    image: ./media/av1-2019-year-in-review.png
date: "2020-01-29"
draft: false
images:
[...]

PR #1389, opened 2024-01-10
support relative paths for opengraph images

chkuendig:

This allows referencing images in page bundles without having to put the whole path into the front-matter (which would break if the page moves or gets a custom slug).

This came up when opengraph template was originally implemented ( https://github.com/adityatelange/hugo-PaperMod/issues/30 and https://github.com/adityatelange/hugo-PaperMod/issues/13, especially https://github.com/adityatelange/hugo-PaperMod/issues/30#issuecomment-706725646 )
## PR Checklist

- [ ] ~This change adds/updates translations and I have used the [template present here](https://github.com/adityatelange/hugo-PaperMod/wiki/Translations#want-to-add-your-language-).~
- [ ] ~This change adds a Social Icon which has a permissive license to use it.~

### DIFF
--- a/layouts/partials/templates/opengraph.html
+++ b/layouts/partials/templates/opengraph.html
@@ -11,7 +11,7 @@
 {{- else }}
 
 {{- with $.Params.images -}}
-{{- range first 6 . }}<meta property="og:image" content="{{ . | absURL }}" />{{ end -}}
+{{- range first 6 . }}<meta property="og:image" content="{{ index (findRE "^/.*" . 1) 0 | default (print $.RelPermalink .) | absURL }}"/>{{ end -}}
 {{- else -}}
 {{- $images := $.Resources.ByType "image" -}}
 {{- $featured := $images.GetMatch "*feature*" -}}


adityatelange:
Relative paths are not recommended to be used with OpenGraph images, they won't work. Give this a read https://ogp.me/#data_types

chkuendig:
I understand, but this isn't what this PR does. It in fact converts relative paths to absolute URLs. This is purely about about accepting relative locations for the assets. 

This is already discussed in detail in the linked above issues (including comments from you).

adityatelange:
Oh my bad I didn't pay enough attention it seems. Those issues are very old and I don't remember much. If you have some free time, can you please share some examples maybe a post.md where we have front-matter? Thanks

chkuendig:
Sure, no worries :) 

To explain in detail what this PR does, let's take the following post from my own page:  https://christian.kuendig.info/posts/2024-01-scummvm-part3/ 

with the PR this is the current situation:
* fontmatter (with the image in the same page bundle folder):
  ```
   images:
     - scummvm-ft-demo-00003.png
  ```
*  output: (I haven't setup a `baseURL` which is why it's missing here )
      ```
      <meta property="og:image" content="/posts/2024-01-scummvm-part3/scummvm-ft-demo-00003.png"/>
      ```

Before this PR, the image URL would have been `/scummvm-ft-demo-00003.png`  which would not work and 404. The only way to make this work currently is to set the full path in the font matter like, the following, but that is very unelegant and breaks if you move the page or it's custom slug at any point without adjusting the font matter.
```
images:
  - /posts/2024-01-scummvm-part3/scummvm-ft-demo-00003.png
```

This is also an issue with the Hugo internal templates, see for example https://discourse.gohugo.io/t/links-for-images-videos-audio-in-internal-templates-twitter-cards-and-opengraph/10464/4 (2018, maybe it's been fixed by now).

You mention in https://github.com/adityatelange/hugo-PaperMod/issues/13#issuecomment-705313991 that this can't be fixed in papermod as it's an issue with the internal hugo templates, but the discussion later mentions various potential fixes which is what this PR is based on:

* current url generation:`{{ . | absURL }}` - this just converts the input string to an absolute url based on the `baseURL`, see https://gohugo.io/functions/urls/absurl/.

* with this PR: `index (findRE "^/.*" . 1) 0 | default (print $.RelPermalink .) | absURL` which works as follows:
  - first checks if the string starts with a slash (i.e.an absolute path) and then passes the whole string if it does or nothing in case it doesn't (`index (findRE "^/.*" . 1) 0`). 
  - if nothing was passed (i.e. a relative path), the next part `default (print $.RelPermalink .) ` takes the original string and gets the relative permalink for it (this is where the relative file paths get converted).
  - and finally we pass it to `absURL` as before the PR to get the full absolute URL. 

This means that absolute paths skip the middle stage and the whole thing works the same as before, only for paths starting without a slash (i.e. relative) the middle pipe is run.

adityatelange:
Understood, but are there any benefits of not setting a baseURL?

chkuendig:
No, thats just a personal thing I forgot to set and wanted to mention why its missing in the example above. 

Doesnt matter when it comes to this PR and this PR doesnt change that as it still runs through `absURL` as it did before so the `baseURL` will still be added if its set. The PR is purely about being able to add a image to the fontmatter with a relative path instead of absolute.

adityatelange:
I am trying to figure this out but I am not getting it. With cover images we use

```yml
# content/blog/mypost/index.md
---
cover: { image: media/cover/secure-http-headers.jpeg, relative: true }
---
```

Image is at `content/blog/mypost/media/cover/secure-http-headers.jpeg`

and it links well without setting the full path here. 

https://github.com/adityatelange/hugo-PaperMod/blob/b5f7118d826e663bfe76f56eba2baa028a384325/layouts/partials/templates/opengraph.html#L5-L10

adityatelange:
Rest of the code is from Hugo templates unmodified.

https://github.com/gohugoio/hugo/blob/cb98e9061b3d62b6f1e635a6cf9a8be57dc3f56c/tpl/tplimpl/embedded/templates/opengraph.html

adityatelange:
There was some development on this https://github.com/gohugoio/hugo/pull/11688#issuecomment-1835416580

and I think https://github.com/gohugoio/hugo/commit/14d85ec136413dcfc96ad8e4d31633f8f9cbf410 fixes the issue?

chkuendig:
Actually that works when using a cover image, but if I don't want to use a cover image but still have an opengraph image, there's a backup path to set it a few lines after:
https://github.com/adityatelange/hugo-PaperMod/blob/b5f7118d826e663bfe76f56eba2baa028a384325/layouts/partials/templates/opengraph.html#L14

This one unfortunately doesn't have the check for a relative path. This is also broken in the original hugo template - which is what I mentioned in the previous comment.

I guess an alternative solution would be to handle this the same way as the cover image (i.e. introduce a separate property to specify whether it's a relative path), but that wouldn't be backward compatible (or we'd still have to add yet another path to check if the image is a object/dict or a string).

adityatelange:
If that fix works, we can clone https://github.com/gohugoio/hugo/commit/14d85ec136413dcfc96ad8e4d31633f8f9cbf410#diff-6be91eea11ef02863538c2b6698248757ad498326c234fe7c085c134c9a4245a to `layouts/partials/templates/_funcs/get-page-images.html` in PaperMod and pick remaining changes for `opengraph` and `twitter_cards` 
- https://github.com/gohugoio/hugo/commit/14d85ec136413dcfc96ad8e4d31633f8f9cbf410#diff-fa08dcd44528f7a509376d58270909ca26c568e1b51e4d6f0b7a77f8cc32e5cd
- https://github.com/gohugoio/hugo/commit/14d85ec136413dcfc96ad8e4d31633f8f9cbf410#diff-5b97abb8773a42aee840ae8559e2961cef2412c3065e2d85778eb06868d74297

chkuendig:
Great catch, I must have missed this (or it wasnt yet done when I looked into it originally).

I think this might solve the issue. Ill check and if it does close this PR and create a superseeding one with the upstream changes if thats okay? I guess we’d have to copy everything to make sure this also works on older versions of hugo, right?

adityatelange:
> Great catch, I must have missed this (or it wasnt yet done when I looked into it originally). I think this might solve the issue. Ill check and if it does close this PR and create a superseeding one with the upstream changes if thats okay? I guess we’d have to copy everything to make sure this also works on older versions of hugo, right?

Yeah, I would have pulled this patch at some point anyways. If you create a PR make sure to add proper authorship for the commits, lets give credits to the person who wrote it :)

PR #1400, opened 2024-01-26
configurable entry summary length

vindex10:

Summaries in the entry list are shortened, with hardcoded number of visible rows.

Even though `summaryLength` can be extended, the visible part in the entry list doesn't change.

In this PR I suggest a way to make it configurable through the template params in `hugo.toml`.

It should be noted, that this PR is blocked by another pull request, which allows parametrizing theme styles: https://github.com/adityatelange/hugo-PaperMod/pull/1399

... and should be rebased when the 1399 is merged.

no

## PR Checklist


adityatelange:
Why not just add this to custom css https://github.com/adityatelange/hugo-PaperMod/wiki/FAQs#bundling-custom-css-with-themes-assets

```css
.entry-content {
    -webkit-line-clamp: <any value>;
}
```

wrobelda:
> ```css
> .entry-content {
>     -webkit-line-clamp: <any value>;
> }
> ```

I guess because the current solution goes against the grain, being completely confusing and incompatible with Hugo's own setup for limiting the length? 

I spent past 40 minutes trying to understand what's causing my content truncated, despite everything set up accordingly. Only after disabling styling one-by-one in Developer mode in Safari I stumbled upon "webkit-line-clamp: 3;".

adityatelange:
> I guess because the current solution goes against the grain, being completely confusing and incompatible with Hugo's own setup for limiting the length?
> 
> I spent past 40 minutes trying to understand what's causing my content truncated, despite everything set up accordingly per Hugo's docs. Only after disabling styling one-by-one in Developer mode in Safari I stumbled upon "webkit-line-clamp: 3;".

I spent more than 40 mins of time debugging, answering queries for this project.

You cannot restrict number of lines with hugo's ways of limiting length. There are devices of different screen width and on all of them you need the exact same heights of cards you see on home page. Guess how would you do it? By line-clamping.

Spend some time thinking why things are like the way they are instead of telling us importance of your time.

wrobelda:
I am letting you know how hard it was to get to understand what’s happening here _as a feedback_. You instigating I consider my time important here is just your projection and I don’t find that acceptable.

I get that you can’t adjust the number of lines, but one would expect a quirk like this to be documented, especially since you arbitrarily set that number to 3. Author of this PR offered to make this configurable and, as such, _explicit_, but you rejected it, throwing people like myself under a bus. And when I explain how difficult it is to get to the bottom of this, you get offensive.

adityatelange:
@wrobelda you don't get to ask me what to do. You don't like it you don't use it. You're feedback lacks appreciation and is just a rant.
Don't expect me to document stuff for you. You're on your own in this foss world. This is a Free software, forkoff and modify it as you like it.

PR #1401, opened 2024-01-26
render short summaries into HTML

vindex10:

It introduces a new config option: ` onlyRenderedShortSummaries: true/false`.

When set to `true`, it renders the summaries in the entry lists. See example on the screenshots.

`.Summar` by default strips html, so the logic is the following:

1. If `.Page.params.summary` is defined in the page frontMatter -> render this summary
2. Else if, `.Summary` was not truncated (`.Truncated` is false) -> render the page `.Contents` for summary.
3. Else, do not show summary at all.

Which means, for the long posts, if summary has to be truncated AND it was not provided in the frontMatter, it will not be shown.

image
image
image
image

additionally, due to rendered HTML (with links) can confusingly lead to the post instead of the link url,
this PR brings the entry summary on top of the link, when "render small summaries" is turned on.

this is how hover on "black" link looks like in this PR:

image

It should be noted, that this PR is blocked by another pull request, which allows parametrizing theme styles: https://github.com/adityatelange/hugo-PaperMod/pull/1399

... and should be rebased when the 1399 is merged.

no

## PR Checklist


adityatelange:
This make the card more confusing for the user. It is better the way it is now.

PR #1408, opened 2024-01-30
add htmx option

lucasew:

I've used turbolinks previously to give that SPA feel to a static site but turbolinks is basically dead.

Today we can achieve the same behaviour using hx-boost from HTMX. This PR adds a simple option to enable that
and a overridable partial to override where the site gets their HTMX bundle.

Vendoring is possible if desirable as HTMX has a permissive enough license to embed without even needing attribution.

## PR Checklist


adityatelange:
HI this is not in scope for papermod

PR #1411, opened 2024-02-01
Adding xmlns:content namespace to RSS feed and only write content:encoded if something on .Content

DJLink:

2 changes to layouts/_default/rss.xml

- Adding a xmlns:content namespace to produce correct RSS output when <content:encoded> is used, namespace from here: https://web.resource.org/rss/1.0/modules/content/

- Added a check to {{- if site.Params.ShowFullTextinRSS }} to also check if .Content, this way pages that don't have anything don't end up with empty content, or for example RSS or tags pages or categories where they don't have .Content

No

## PR Checklist


adityatelange (approved):
Thanks :+1:

PR #1439, opened 2024-02-27
Separate post-footer for better customization

Beryesa:

Separates post-footer into post_footer.html for better customization
Leaves less room for git conflicts and unnecessary overwrites, which results in better flexibility

Not really, but following post-meta for consistency, which is also separate 

## PR Checklist


adityatelange:
post-meta is separate because we call it from multiple template, with post-footer it is only on one template.

PR #1494, opened 2024-04-17
Remove the use of accesskey

xuhdev:

`accesskey` is quite problematic. MDN advises not using accesskeys for general purpose websites for various reasons: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/accesskey#accessibility_concerns

WebAIM also recommends against using accesskeys generally: https://webaim.org/techniques/keyboard/accesskey

No.

## PR Checklist

### DIFF
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -23,7 +23,7 @@
 {{- end }}
 
 {{- if (not site.Params.disableScrollToTop) }}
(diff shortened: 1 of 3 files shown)


adityatelange:
It will stay until browsers remove the support

xuhdev:
I don't think browsers will ever drop support because access keys are useful in some cases. The PR description says they are not recommended to be used in a general website, citing various sources.

adityatelange:
image
Some users find this useful

PR #1498, opened 2024-04-21
Added umami analytics support

DJLink:

This PR adds umami analytics support (https://umami.is/), the user can add the id and url of their installation to analytics and activate the tracking code.

it can be achieved with 

```
params:
      analytics:
          umami:
              id: ID_CODE  #provided on umami backoffice
              url: URL  #for example https://eu.umami.is/script.js

````

This will add the appropriate code as it can be viewed on my website which is already using this change live:

https://github.com/DJLink/web_david-amador.com/blob/9234c4cfb7935c224abc123dad83a59ff7a7ec5e/index.html#L31

No

## PR Checklist

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -151,6 +151,7 @@
 {{- /* Misc */}}
 {{- if hugo.IsProduction | or (eq site.Params.env "production") }}
 {{- template "_internal/google_analytics.html" . }}
+{{- template "partials/templates/umami.html" . }}
 {{- template "partials/templates/opengraph.html" . }}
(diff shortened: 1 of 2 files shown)


adityatelange (commented):
Umami gives users a pre-generated HTML code to paste into any template.
image

I as a user would simply make copy the provided snippet into `extend_footer.html` https://adityatelange.github.io/hugo-PaperMod/posts/papermod/papermod-faq/#custom-head--footer

Also umami allows more customization and data parameters https://umami.is/docs/tracker-configuration ,we can't cover all that here.

PR #1499, opened 2024-04-25
feat: remove json index on search-hidden pages

Singularity0909:

To remove JSON index on the page set search-hidden param as true.

No.

## PR Checklist

**I'm not sure about the last check item.**

### DIFF
--- a/layouts/partials/templates/schema_json.html
+++ b/layouts/partials/templates/schema_json.html
@@ -55,7 +55,7 @@
   ]
 }
 </script>
-{{- if .IsPage }}
+{{- if (and (.IsPage) (not .Params.searchHidden)) }}
 <script type="application/ld+json">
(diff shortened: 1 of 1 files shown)


adityatelange (commented):
How does this affect Search? json schema has no effect on search.

Singularity0909:
> How does this affect Search? json schema has no effect on search.
> 
> 
> 
> For context searchHidden is to hide pages from site's search index and not for any search engine's 

Thanks for reviewing. Sometimes I encrypt some pages' partial content of my blog.

https://github.com/Li4n0/hugo_encryptor

Through this tool, some parts of an article will be encrypted with AES and encoded into base64. However, JSON of the entire original content in the header of the webpage cannot be encrypted, which may be spied on.

PR #1532, opened 2024-05-31
When display post summary, use .Summary instead of .Description

xuhdev:
.Description is meant to be used for metadata purpose: https://gohugo.io/methods/page/description/
While .Summary can be used to display to human readers: https://gohugo.io/methods/page/summary/

Hugo document has emphasized this in the .Description page:

> Conceptually different from a content summary, a page description is typically used in metadata about the page.

Hugo embedded templates are consistent in this regard.

## PR Checklist

### DIFF
--- a/assets/css/common/post-single.css
+++ b/assets/css/common/post-single.css
@@ -8,7 +8,7 @@
     font-size: 40px;
 }
 
-.post-description {
+.post-summary {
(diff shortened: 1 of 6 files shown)


xuhdev:
Patch seems to be malfunctioning, closing

PR #1581, opened 2024-08-25
feature:(svg): Change the fediverse symbol to ⁂

bocan:

As proposed on the [fediverse.info page](https://symbol.fediverse.info/), how about using the asterism symbol (**⁂**) as a new fediverse icon?  I've swapped it in to see what people think.  I'm not sure whether I prefer it myself, but whether you want to merge it or not, it could be useful if someone likes it better.

As noted in the link above, the SVG image is in the public domain.

It ends up looking like the middle icon here:
Screenshot 2024-08-25 at 02 47 40
Screenshot 2024-08-25 at 02 47 49

No.

## PR Checklist

- [x] This change adds a Social Icon which has a permissive license to use it.


adityatelange:
The new icon is a proposal but I'm not sure if it's being used yet. Most of them use the current icon which you can see on wikipedia. Having an icon already known by people will be better until it's popular enough to get replaced (similar to twitter -> x).
Closing for now, we'll merge if it gets mainstream. Thanks!

PR #1594, opened 2024-09-07
Update he.yaml

idanhazan:
### This PR includes three types of changes:

- **Completion of Missing Translations:** Added translations for previously missing content.
- **Content Corrections:** Corrected translations to better reflect the intended meaning.
- **Language Direction Adjustments (RTL):** Fixed issues related to the right-to-left (RTL) text direction to ensure proper display and readability.

Please review the changes and let me know if there are any additional adjustments needed. Thank you!

### DIFF
--- a/i18n/he.yaml
+++ b/i18n/he.yaml
@@ -1,5 +1,5 @@
 - id: prev_page
-  translation: "קודם"
+  translation: "הקודם"
 
 - id: next_page
   translation: "הבא"
@@ -7,15 +7,15 @@
 - id: read_time
   translation:
     one : "דקה אחת"
-    other: "דקות {{ .Count }}"
+    other: "{{ .Count }} דקות"
 
 - id: words
   translation:
-    one : "word"
-    other: "מילים {{ .Count }} "
+    one : "מילה"
+    other: "{{ .Count }} מילים"
 
 - id: toc
-  translation: "תוכן העניינים"
+  translation: "תוכן עניינים"
 
 - id: translations
   translation: "תרגומים"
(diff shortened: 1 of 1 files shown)


adityatelange:
@eisraeli @RealA10N requesting your review

idanhazan:
I had two additional questions:

1. Should we use "דקה 1" instead of "דקה אחת" to maintain consistency with the English version, which uses "1 min" rather than "one min"?

2. Should we use "מילה 1" or "מילה אחת" instead of just "מילה" to align with the format used for minutes?

RealA10N:
IMO there is no need to enforce consistency across languages at this micro level: as long as the meaning is consistent, the exact wording should be in accordance with the laws of the local language.

So I suggest:
> מילה אחת
For "1 word", and,
> דקה אחת
For "1 min".

Other than that, LGTM!

i18n/he.yaml
```diff
@@ -1,21 +1,21 @@
...
   translation:
     one : "דקה אחת"
-    other: "דקות {{ .Count }}"
+    other: "{{ .Count }} דקות"
 
 - id: words
   translation:
-    one : "word"
-    other: "מילים {{ .Count }} "
+    one : "מילה"
```
RealA10N, line 14: I suggest

> מילה אחת

This phrasing is also consistent with `read_time.one`

RealA10N (approved):

i18n/he.yaml
idanhazan: Done ✅

PR #1597, opened 2024-09-14
feat: limits how many posts will be listed on the home page

Lkzeu:
#### Limits how many posts will be listed on the home page.
This PR creates a new int variable called `maxPostsOnHomePage` on `site.Params` that will limit how many posts will be listed on the home page.

To use it, just add the variable in hugo.yaml as below
```
params:
    maxPostsOnHomePage: 3 # any integer, 0 = unlimited.
```

Once set, it will only list the  _N_ most recent posts.
I tested it with Regular and Home-Info modes and it did work as expected.

## PR Checklist

### DIFF
--- a/layouts/_default/list.html
+++ b/layouts/_default/list.html
@@ -43,6 +43,10 @@
 {{- if .IsHome }}
 {{- $pages = where site.RegularPages "Type" "in" site.Params.mainSections }}
 {{- $pages = where $pages "Params.hiddenInHomeList" "!=" "true"  }}
+{{- if site.Params.maxPostsOnHomePage }}
+{{- $limit := site.Params.maxPostsOnHomePage }}
+{{- $pages = $pages | first $limit }}
+{{- end }}
 {{- end }}
 
 {{- $paginator := .Paginate $pages }}


adityatelange:
https://github.com/adityatelange/hugo-PaperMod/blob/2378012732ae821b7f987a941919cbb06b25a0d5/config.yml#L6

Lkzeu:
> https://github.com/adityatelange/hugo-PaperMod/blob/2378012732ae821b7f987a941919cbb06b25a0d5/config.yml#L6

Thank you for your reply, @adityatelange!

`paginate: 5` will make each page on your site only have 5 items. If the page has more, a “next” button will be placed at the end of the listed items.

but what if I want to list 10 posts in `/posts/` but only 3 on my home page? Also, I don't want "next" to appear on the home page. I could put a "see all posts" at the end of the list that will point to `/posts/` and so on.

If `maxPostsOnHomePage <= paginate` then it will limit the number of posts shown on the home page without a "next" button and without changing how items are listed on other pages of the site, making this theme even more customizable.

---
Anyway, great theme and great work!
Thanks in advence.

PR #1605, opened 2024-09-20
Bump GitHub workflow actions to latest versions

deining:
This PR bump GutHub workflow actions to latest versions, thus avoiding deprecation warnings as seen e.g. [here](https://github.com/adityatelange/hugo-PaperMod/actions/runs/10869399986).


adityatelange:
This looks already pushed last week. Thanks btw.

Issue #1612, opened 2024-10-07
[Enhancement]: Optional special handling of external links


David-Guillot:
### What you'd like to propose?

As @DanilaFe stated in [this blog post](https://danilafe.com/blog/blog_microfeatures/#markers-for-external-links) (which got quite popular, see on [lobsters](https://lobste.rs/s/ju9yby/) and [hn](https://news.ycombinator.com/item?id=40774277)), marking external links is helpful when reading a blog. Maybe this Hugo theme can do something about it.

My suggestion is:
- One Hugo setting that enables opening external links to a new tab
- One Hugo setting that enables a small icon at the end of an external link label

What do you think?

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1613, opened 2024-10-07
[FEATURE] optional special treatment for external links

David-Guillot:

This PR proposes a way to optionally open external links to a new tab, and optionally mark external links with an icon.

I just opened #1612 to make a rationale about this.

## PR Checklist

### DIFF
--- a/assets/css/common/post-single.css
+++ b/assets/css/common/post-single.css
@@ -82,6 +82,46 @@
     box-shadow: 0 -1px 0 var(--primary) inset;
 }
 
+.post-content a.external-link {
+  display: inline-flex;
+  align-items: center;
+  gap: 5px;
+}
+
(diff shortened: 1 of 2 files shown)


adityatelange:
We already do  this with external links in menu https://adityatelange.github.io/hugo-PaperMod/
image

David-Guillot:
> We already do this with external links in menu https://adityatelange.github.io/hugo-PaperMod/ image

Ah cool! So how do you feel about this PR? It brings two differences:

- The icon is pure CSS, no SVG involved, so it's more lightweight and with better accessibility (screen readers won't try to read the image)
- The ability to open external links in a new tab is added through a site-level option

Let me know if you think this should be implemented another way. Or if you think this implementation should be applied to what already exists in the header menu.

PR #1619, opened 2024-10-17
[FEATURE] new opengraph 'fediverse:creator' meta

### LINKED ISSUE #1618: [Enhancement]: HTML meta "fediverse:creator"
### What you'd like to propose?

This summer Mastodon [introduced](https://blog.joinmastodon.org/2024/07/highlighting-journalism-on-mastodon/) a feature that allows to highlight the author of an article/post share on the Fediverse. It's been released this week with Mastodon 4.3 which has been deployed on most major instances, and it's awesome because it increases the visibility of everything we write and it allows people to easily browse contents from people who inspire them.

This feature relies on a new HTML OpenGraph meta tag:

```html
<meta name="fediverse:creator" content="@user@instance" />
```

It would be pretty easy for this Hugo theme to support this new OpenGraph tag, based on a new configuration setting that would be fetch from a post front matter and would default on a site param. I'll open a PR right away.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

David-Guillot:

This PR adds support of [this new Mastodon 4.3 feature](https://blog.joinmastodon.org/2024/07/highlighting-journalism-on-mastodon/) for highlighting journalists and bloggers.

It does so by adding a new OpenGraph meta based on a new param which can be placed on a post or on the Hugo site.

Closes #1618 #1642

**Usage**

Site level:
```yml
params:
  social:
    fediverse_creator: "@adityatelange@mastodon.social"
    twitter: adityatelange
```
Page (Overrides Site level value):

```yml
---
title: "My Page"
social:
  fediverse_creator: "@adityatelange@mastodon.social"
---
```

## PR Checklist

### DIFF
--- a/layouts/partials/templates/opengraph.html
+++ b/layouts/partials/templates/opengraph.html
@@ -61,6 +61,10 @@
   {{- end }}
 {{- end }}
 
+{{- if or .Params.fediverseCreator site.Params.fediverseCreator }}
+<meta name="fediverse:creator" content="{{ .Params.fediverseCreator | default site.Params.fediverseCreator }}" />
+{{- end -}}
+
 {{- range .GetTerms "series" }}
   {{- range .Pages | first 7 }}
     {{- if ne $ . }}
--- a/layouts/partials/templates/opengraph.html
+++ b/layouts/partials/templates/opengraph.html
@@ -61,10 +61,6 @@
   {{- end }}
 {{- end }}
 
-{{- if or .Params.fediverseCreator site.Params.fediverseCreator }}
-<meta name="fediverse:creator" content="{{ .Params.fediverseCreator | default site.Params.fediverseCreator }}" />
-{{- end -}}
-
 {{- range .GetTerms "series" }}
   {{- range .Pages | first 7 }}
     {{- if ne $ . }}
@@ -83,4 +79,8 @@
       {{- end }}
     {{- end }}
   {{- end }}
+{{- end }}
+
+{{- with (.Param "social.fediverse_creator") }}
+  <meta name="fediverse:creator" content="{{ . }}">
 {{- end }}


adityatelange:
Hi @David-Guillot I think its better if we add it to hugo's templates over here. https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/opengraph.html

cc @jmooring

jmooring:
This would be another opinionated change to one of Hugo's embedded templates; I am not in favor of doing this. If it were up to me I would remove Hugo's embedded opengraph template (and others) from the code base.

David-Guillot:
@adityatelange I think I agree with you: this theme is about styling a Hugo blog, not adding features that could apply to all Hugo bolgs. I just didn't think of opening a PR to @gohugoio/hugo .

@jmooring , are you saying that porting this PR to @gohugoio/hugo is a bad idea, because it would make Hugo too specific about what the web landscape is?

How do we move forward from here?

jmooring:
I'm saying that this should be implemented at the theme or site level.

adityatelange:
> I would remove Hugo's embedded opengraph template (and others) from the code base.

Think not of like a hugo author but a theme author. we have a feature dependency here with cover images which we cannot hot-plug into hugo's templates. 

I had to overwrite embedded templates because we don't have a way to add hooks to these templates. If you have a better way of doing this please suggest. For hugo this is not a hurdle because the feature we are talking about is also a opinionated feature here.

adityatelange:
@David-Guillot 
>

jmooring:
In my view many of the embedded templates are, at best, opinionated examples, and _should_ be overwritten by theme authors to achieve the desired outcome.

Over the years I've examined thousands of sites, and I have seen all of the configuration constructs below:

```text
[params]
fediverseCreator = 'foo'

[params.social]
fediverseCreator = 'foo'

[params.social.fediverse]
creator = 'foo'

[[params.social]]
id = 'fediverse'
creator = 'foo'

[[params.social.networks]]
id = 'fediverse'
creator = 'foo'

[[params.mytheme.social.networks]]
id = 'fediverse'
creator = 'foo'
```

Is one of these better than the others? No; it depends on what the site or theme author has in mind.

jmooring:
> but a mapping of each authors fediverse id

Yet another reason for handling this at the theme or site level.

adityatelange:
> > but a mapping of each authors fediverse id
> 
> Yet another reason for handling this at the theme or site level.

True

David-Guillot:
> I see we also support multiple authors in meta tags, so in order to have this supported we'll not just need one name but a mapping of each authors fediverse id. Is it feasible? idk

I'll try to figure that out next week. Meanwhile I'll convert this PR to draft. Thank you both for the discussion!

David-Guillot:
@adityatelange for now, as you can see in [The Technical section](https://blog.joinmastodon.org/2024/07/highlighting-journalism-on-mastodon/#the-technical) of Mastodon's announcement, *multiple authors are not yet supported on their end**.

So I just rebased this PR for now, let me know if you want to integrate this change as-is, in which case I'll set the PR ready for review.

Syphdias:
The mentioned article is from 2024-07-02 says:

>If multiple tags are present on the page, the first one will be displayed, but we may add support for showing multiple authors in the future.

This is very vague and might be never. The following sentence about "coming weeks" is not about multiple authors but about the single creator attribution they just announced for Mastodon.

>We intend to propose a specification draft for other ActivityPub platforms in the coming weeks.

I think there is no use on waiting for a new spec unless there is already a draft of the spec with multiple author attribution somewhere.

David-Guillot:
@adityatelange what's your thoughts on this? I think @Syphdias is right: multiple authors are nowhere near landing on Mastodon. But many bloggers would benefit from this PR.

adityatelange:
> @adityatelange what are your thoughts on this, two months later? I think @Syphdias is right: multiple authors are nowhere near landing on Mastodon. But many bloggers would benefit from this PR.

Yes. #1642 looks much simpler patch to me. We can go ahead with this.

@jmooring would you be open to integrate #1642 to hugo's internal templates as well. Since we have `fb:app_id` and `fb:admins` which are also similar proprietary meta tags unrelated to [opengraph protocol](https://ogp.me/).

David-Guillot:
Yeah I don't see much difference between #1642 and this one (this one just allows defining an author at site-level and one at post-level, which I thought was a good idea), but it you prefer #1642 let's go for it, because I really think this feature is a must-have for many users!

adityatelange:
Right that seems to be missing, I'll add something for that

adityatelange:
Usage:

Site level:
```yml
params:
  social:
    fediverse_creator: "@adityatelange@mastodon.social"
    twitter: adityatelange
```
Page (Overrides Site level value):

```yml
---
title: "My Page"
social:
  fediverse_creator: "@adityatelange@mastodon.social"
---
```

PR #1621, opened 2024-10-18
wrap highlight code

mrasong:


adityatelange:
Thank you for your feature request! After careful consideration, we've decided not to implement this feature at this time as it doesn't align with our current priorities. We appreciate your understanding and encourage you to share any other ideas you may have!

PR #1626, opened 2024-11-03
i18n chinese for titles and breadcrumbs of series and tags

liupras:
When the default language is set to zh, the titles and breadcrumbs of /tags and /series are still in English.
So I modified i18n/zh.yaml, layouts/_default/terms.html, layouts/partials/breadcrumbs.html so that the titles and breadcrumbs of /tags and /series can be displayed in Chinese.

### DIFF
--- a/i18n/zh.yaml
+++ b/i18n/zh.yaml
@@ -31,3 +31,12 @@
 
 - id: code_copied
   translation: "已复制！"
+  
+- id: articles
+  translation: "所有文章"
+
+- id: Series
+  translation: "合集"
+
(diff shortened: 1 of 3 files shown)


i18n/zh.yaml
```diff
@@ -31,3 +31,9 @@
 
 - id: code_copied
   translation: "已复制！"
+  
+- id: Series
```
jsolivellase, line 35: By default, Hugo only defines tags and categories as taxonomies. I think we should translate tags and categories only.

PR #1629, opened 2024-11-15
add share icon for bluesky

geowarin:

Adds a share icon for bluesky

The social icon was added but not the share button

## PR Checklist

- [x] This change adds a Social Icon which has a permissive license to use it.


adityatelange:
we cant have a single domain for a decentralised platform. you can't make it customisable because you need to ask user where he/she want to post it.

geowarin:
Well it is not federated yet (that I know of). 
Maybe we can start with that and re-evaluate when there are other instances (but I believe most people will stay on the same instance)

siddharthgoel88:
@adityatelange plan to release this soon?

PR #1638, opened 2024-11-23
Add a share button for bluesky

newptcai:

I added a share icon for bluesky.
In my test it works for both light and dark mode.
The icon comes from https://simpleicons.org/?q=bluesky
I modified it so it is compatible with other share icons.
It has Creative Commons Zero v1.0 Universal (CC0 1.0) license.
So there's no problem using it.

No.

## PR Checklist


adityatelange:
Dup #1629

PR #1660, opened 2025-01-14
Improve some terms for Bengali translation

tarunjana:

This PR updates some terms in Bengali translation for better readability.

No

## PR Checklist


adityatelange:
@ayonkhan can you please help with reviewing this?

ayonkhan:
Thanks for the PR and the work put into updating the Bengali translations. After reviewing the changes, I wanted to share my thoughts on the three suggestions.

First, for the `one: "1 min"` translation, the current Bengali version is `one: "১ মিনিট"`, which keeps the numeric representation, just like the English version. I think this works well because it's consistent with the original. The suggested change, replacing the number `1` with the word `এক`, doesn't seem quite right, as it shifts away from the numeric format in the English version. I feel it's important to keep that numeric consistency in the Bengali translation.

Next, for the `one` and `other` keys under the ID `words`, the current translation is `one: "১ টি শব্দ"` and `other: "{{ .Count }} টি শব্দ"`. Both use numeric representations, keeping them consistent with each other. The new suggestion, however, replaces the number in `one` key with words (e.g., `একটি শব্দ`), while the `other` key still uses the numeric format. This introduces an inconsistency between the two. I believe it would be better to keep the numeric format in `one` to ensure both keys are aligned.

Lastly, for the third suggestion regarding `সূচিপত্র` vs `বিষয়সারণী`, I think `সূচিপত্র` is more commonly used and sounds more natural in everyday conversation. While `বিষয়সারণী` carries the same meaning, it's a bit of a mouthful and less commonly heard. That said, I understand this is a subjective choice, and I'm open to further discussion.

Thanks again for considering my feedback, and I'm happy to talk more about these points if needed!

tarunjana:
Thank you for the detailed explanation. I completely agree that the changes are a little bit subjective. So I am closing this.

Issue #1716, opened 2025-04-08
[Enhancement]: Post list links should render as links in text-based browsers


jackbaty:
### What you'd like to propose?

I realize I'm in a small minority, but I sometimes browse the web using w3m in Emacs.

When using w3m, posts listed on the home page do not render a link to each post. Could perhaps the title (in `<h2 class="entry-hint-parent">`) also include an explicit link?

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

gvisoc:
I think this is a bug, rather than an enhancement. The blogs aren't navigable at all on a terminal without those links; while I am not an accessibility expert I bet this also breaks navigation driven by screen readers.

PR #1725, opened 2025-04-14
support links in text-based browsers

SeanMooney:
This change modifies the list template to add
an additional <a> to the article header.

The existing a tag uses the `entry-link` css class
to make the containing article cliable by forcing
`position: absolute` in combination with an empty
anchor element `<a class='entry-link'></a>`

This has the effect of making the entire article
clickable to navigate in a graphical browser however
it is not rendered in a text-based browser.

To correct this the title text in the article header
is wrapped in a second <a> tag. The `entry-link` css
class is not applied and default css is inherited.

Closes: #1716

## PR Checklist

### DIFF
--- a/layouts/_default/list.html
+++ b/layouts/_default/list.html
@@ -68,7 +68,8 @@
   {{- partial "cover.html" (dict "cxt" . "IsSingle" false "isHidden" $isHidden) }}
   <header class="entry-header">
     <h2 class="entry-hint-parent">
-      {{- .Title }}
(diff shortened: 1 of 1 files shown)


SeanMooney:
https://imgur.com/a/w7T32Ct

i have uploaded screenshot of my block using this theme with the master behavior and the behvior form this pr.

this is using the EWW text browser built into emacs to render the site with the local hugo development server

PR #1726, opened 2025-04-15
Implement fixes for refreshed template system for Hugo v0.146.0+

### LINKED ISSUE #1719: Missing template definitions cause production build failure
### What you'd like to propose?

Hi!

In the latest version of PaperMod, the following templates are referenced inside _partials/head.html:

```
{{ template "partials/templates/opengraph.html" . }}
{{ template "partials/templates/twitter_cards.html" . }}
{{ template "partials/templates/schema_json.html" . }}
```

However, these templates are not defined in the theme (using {{ define ... }}), which causes the Hugo build to fail in production mode:

```
Error: html/template:_partials/head.html:155:13: no such template "partials/templates/twitter_cards.html"
```

This is especially confusing because it only triggers in production (i.e., hugo.IsProduction == true), so users upgrading from an older version may not notice the issue until CI breaks.

Suggested Fix:

* Provide empty stub templates for these in layouts/partials/templates/, e.g.:

```
{{ define "partials/templates/twitter_cards.html" }}{{ end }}
```

That way, users can override them if needed, but the build won’t fail by default.

Thanks for the great theme!

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

jakebailey:

Pulls the patch from https://github.com/adityatelange/hugo-PaperMod/issues/1719#issuecomment-2797327929, along with bumping the minimum version. 

Makes Papermod compatible with https://github.com/gohugoio/hugo/releases/tag/v0.146.0

Fixes #1719

## PR Checklist

- [x] ~This change adds/updates translations and I have used the [template present here](https://github.com/adityatelange/hugo-PaperMod/wiki/Translations#want-to-add-your-language-).~
- [x] ~This change adds a Social Icon which has a permissive license to use it.~

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yaml
+++ b/.github/ISSUE_TEMPLATE/bug.yaml
@@ -39,8 +39,8 @@ body:
       label: Hugo Version
       description: What version of Hugo are you running?
       options:
-        - Hugo >= 0.125.7 (Recommended - Minimum version required for PaperMod)
-        - Hugo < 0.125.7  (Incompatible - Not recommended to build PaperMod on lower versions)
+        - Hugo >= 0.146.0 (Recommended - Minimum version required for PaperMod)
+        - Hugo < 0.146.0  (Incompatible - Not recommended to build PaperMod on lower versions)
     validations:
       required: true
   - type: input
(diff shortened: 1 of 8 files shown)


sammcj:
Confirmed this PR seems to fix the related issues.

adityatelange (approved):

PR #1734, opened 2025-04-27
fix/googleanalytics: add dynamic Google Analytics partial using params

prashanthisadde:
Fix: add dynamic Google Analytics partial using params

- Added layouts/partials/google_analytics.html
- Fetch Google Analytics ID from `.Site.Params.services.googleAnalytics.id`
- Load GA script only if environment is 'production' and ID is present
- Updated examplesite/config.yml with sample Google Analytics config (in another PR to examplesite branch)
- Ensures compatibility with Hugo 0.111+ where `.Site.GoogleAnalytics` and `.Site.Services` are deprecated


adityatelange:
Hey, hugo already has support for GA https://gohugo.io/methods/site/config/ Set using .Site.Config.Services.GoogleAnalytics.ID 

> Ensures compatibility with Hugo 0.111+ 

We don't support 0.111 anymore, only 0.146+

PR #1735, opened 2025-04-27
fix-exmpleSite/googleanalytics: add Google Analytics config example under examplesite

prashanthisadde:
docs: add example Google Analytics config in examplesite

- Adds sample Google Analytics configuration under `params.services.googleAnalytics.id`
- Helps demonstrate usage of the new dynamic GA partial
- Continuation of master PR: https://github.com/adityatelange/hugo-PaperMod/pull/1734

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -182,6 +182,11 @@ params:
   # icon: icon.png
   # iconHeight: 35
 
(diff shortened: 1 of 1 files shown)


adityatelange:
related to #1734

PR #1777, opened 2025-08-10
Fix the figure shortcode for multilingual site

sammedhi:

Fix the figure shortcode for multilingual site

The figure shortcode now use RelPermalink instead of the link provided by the user. Using the link provided by the user caused the image not to load when using local url for translated the site was translated. When you translated the site the url is modified (eg: site/fr/post/article/), if you tried to load the image located at "site/post/article/img.png" using {{\<figure src="img.png" \>}} the image would not load since the resulting link would be "site/fr/post/article/img.png". 

This would not work for site using translation by file name https://gohugo.io/content-management/multilingual/#translation-by-file-name

Which is inconsistent with the markdown behaviour where local path works fine.

No, it just happened that the default hugo shortcode was working while the paperMod's one was not loading my image on my translated pages

## PR Checklist

### DIFF
--- a/layouts/shortcodes/figure.html
+++ b/layouts/shortcodes/figure.html
@@ -1,3 +1,5 @@
+
+
 <figure{{ if or (.Get "class") (eq (.Get "align") "center") }} class="
            {{- if eq (.Get "align") "center" }}align-center {{ end }}
            {{- with .Get "class" }}{{ . }}{{- end }}"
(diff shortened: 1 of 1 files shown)


Drumclem:
Hi! Thank you for this PR. I'm having a bit of the same problem with my website (https://clement-martin.fr : the image cards don't work when I'm sharing posts on social media (except for the front page of my website, which fetches the image from /static/). I'm using filename translations (index.fr.md, index.en.md in the same folder for one given post for instance).

Do you have any idea what I should modify to get rid of this bug?

PR #1808, opened 2025-10-24
Use data-theme attribute for theme switching and set color-scheme

taoyouh:

When dark/light mode is selected, not only set the css colors but also set the color-scheme variable, so that native controls (if any, such as custom HTML embedded in markdown) have more natural color by default.

This pull request refactors the site's theme switching logic and CSS selectors to use the `data-theme` attribute on the `html` element instead of toggling the `dark` class on the `body`. This change improves consistency and maintainability for theme-dependent styles, and updates both the JavaScript and CSS to reflect the new approach. The most important changes are grouped below:

**Theme Switching Logic**

* All theme toggle and initialization scripts now set the `data-theme` attribute on the `html` element, replacing the previous use of the `dark` class on the `body`. This affects files such as `header.html` and `footer.html`. [[1]](diffhunk://#diff-42986cf40ad96f5b7ca39d9b71ba6a9f252656de24cfe82100892ff0f4400f90L7-R29) [[2]](diffhunk://#diff-42986cf40ad96f5b7ca39d9b71ba6a9f252656de24cfe82100892ff0f4400f90L36-R40) [[3]](diffhunk://#diff-1236858a4652b74ac20c849dced102285e73793914ce9093a079f5466bb7a938L84-R89)

**CSS Selector Updates**

* CSS selectors throughout the codebase now use `[data-theme="dark"]` and `[data-theme="light"]` instead of `.dark` or `:not(.dark)`, ensuring styles are applied based on the `data-theme` attribute. This includes updates in `theme-vars.css`, `header.css`, `post-single.css`, and `scroll-bar.css`. [[1]](diffhunk://#diff-4f10c9004c75e28fcff636535d72206132b8096cf8bec7ee14f7bd556df6a82eR18-R21) [[2]](diffhunk://#diff-4f10c9004c75e28fcff636535d72206132b8096cf8bec7ee14f7bd556df6a82eR31-R38) [[3]](diffhunk://#diff-ff89aa6413d3427783674195af0ea0ce6300b5efba9768950adbcf6f634d22c1L44-R48) [[4]](diffhunk://#diff-396f1622d834975015b1acae313ca0dc3e8097680eb367da07a18fd8a8ece82cL271-R271) [[5]](diffhunk://#diff-3cc5d54c5f957613676bca56382d9c635fa929cbb1edf0bf7abce5033180eb2aL6-R12)

**Theme Variable Improvements**

* The CSS variables for theme colors and `color-scheme` are now scoped to `[data-theme="dark"]` and `[data-theme="light"]`, further aligning the code with the new theme management approach. [[1]](diffhunk://#diff-4f10c9004c75e28fcff636535d72206132b8096cf8bec7ee14f7bd556df6a82eR18-R21) [[2]](diffhunk://#diff-4f10c9004c75e28fcff636535d72206132b8096cf8bec7ee14f7bd556df6a82eR31-R38) [[3]](diffhunk://#diff-96dd75a968976edd5e03170268ed9085733f75c3fb24f992ae613c89e6de42dcR130-R144)

**Scroll Bar and TOC Styling**

* Scroll bar and table of contents (TOC) styles now rely on the `data-theme` attribute, ensuring correct appearance in both light and dark modes. [[1]](diffhunk://#diff-3cc5d54c5f957613676bca56382d9c635fa929cbb1edf0bf7abce5033180eb2aL6-R12) [[2]](diffhunk://#diff-96dd75a968976edd5e03170268ed9085733f75c3fb24f992ae613c89e6de42dcR130-R144) [[3]](diffhunk://#diff-396f1622d834975015b1acae313ca0dc3e8097680eb367da07a18fd8a8ece82cL271-R271)

These changes make the theme system more robust and easier to extend, and ensure consistent behavior across the site.

Not discussed before.

## PR Checklist

### DIFF
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -81,11 +81,12 @@
 {{- if (not site.Params.disableThemeToggle) }}
 <script>
     document.getElementById("theme-toggle").addEventListener("click", () => {
-        if (document.querySelector("html").getAttribute("data-theme") === "dark") {
-            document.querySelector("html").setAttribute("data-theme", 'light');
+        const html = document.querySelector("html");
+        if (html.dataset.theme === "dark") {
+            html.dataset.theme = 'light';
             localStorage.setItem("pref-theme", 'light');
         } else {
-            document.querySelector("html").setAttribute("data-theme", 'dark');
+            html.dataset.theme = 'dark';
             localStorage.setItem("pref-theme", 'dark');
         }
     })
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -4,7 +4,7 @@
 {{- if (eq site.Params.defaultTheme "light") }}
 <script>
     if (localStorage.getItem("pref-theme") === "dark") {
-        document.querySelector("html").setAttribute("data-theme", 'dark');
+        document.querySelector("html").dataset.theme = 'dark';
     }
 
 </script>
@@ -12,7 +12,7 @@
 {{- else if (eq site.Params.defaultTheme "dark") }}
 <script>
     if (localStorage.getItem("pref-theme") === "light") {
-        document.querySelector("html").setAttribute("data-theme", 'light');
+        document.querySelector("html").dataset.theme = 'light';
     }
 
 </script>
@@ -20,13 +20,13 @@
 {{- /* theme is auto */}}
 <script>
     if (localStorage.getItem("pref-theme") === "dark") {
-        document.querySelector("html").setAttribute("data-theme", 'dark');
+        document.querySelector("html").dataset.theme = 'dark';
     } else if (localStorage.getItem("pref-theme") === "light") {
-       document.querySelector("html").setAttribute("data-theme", 'light');
+       document.querySelector("html").dataset.theme = 'light';
     } else if (window.matchMedia('(prefers-color-scheme: dark)').matches) {
-        document.querySelector("html").setAttribute("data-theme", 'dark');
+        document.querySelector("html").dataset.theme = 'dark';
     } else {
-        document.querySelector("html").setAttribute("data-theme", 'light');
+        document.querySelector("html").dataset.theme = 'light';
     }
 
 </script>
@@ -35,9 +35,9 @@
 {{- else if (and (ne site.Params.defaultTheme "light") (ne site.Params.defaultTheme "dark"))}}
 <script>
     if (window.matchMedia('(prefers-color-scheme: dark)').matches) {
-        document.querySelector("html").setAttribute("data-theme", 'dark');
+        document.querySelector("html").dataset.theme = 'dark';
     } else {
-        document.querySelector("html").setAttribute("data-theme", 'light');
+        document.querySelector("html").dataset.theme = 'light';
     }
 
 </script>


adityatelange (commented):
Doesn't change the browser theme to light/dark.

Use `disableScrollBarStyle: true` disable the scrollbar style so that the browser sets the mode.

taoyouh:
Sorry I don't quite understand. This PR intends to make HTML without custom CSS to have acceptable visibility by default. For example I can add a checkbox in my content:
```
<input type="checkbox" id="contactFormAcceptPolicy" required />
```
With my PR it looks like this:
image

Without my PR the checkbox has inverted color scheme:
image

I don't see any effect of this PR on scroll bar style or browser scheme.

taoyouh:
Sorry wrong image attached. I've updated my comment above.

adityatelange:
Right, but if we are setting `color-scheme` it should change color of all browser-provided UI elements including scrollbars. This way we affect only internal elements inside the body tag as `.dark` is added to body only.

taoyouh:
I assumed if the body is using dark scheme, scrollbars and other browser elements inside body should also be dark scheme for consistency. If we ever use native scrollbar, a dark-scheme page with ligh-scheme scroll bar shall look weird.

Do you have some other consideration on that? Or, will it look better if I set the color-scheme on `body.dark` instead of `:root`?

adityatelange:
@taoyouh 
Done some major changes for color-scheme to work with/without js enabled. I think this would work
Also if you see any other color issues, please let me know

assets/css/common/post-single.css
```diff
@@ -268,7 +268,7 @@
     padding: 0.4em;
 }
 
-.dark .toc {
+[data-theme="dark"] .toc {
```
taoyouh: We can use `background: light-dark(var(--code-bg), var(--entry))` instead to avoid setting the same thing twice in both `[data-theme="dark"]` and `@media (prefers-color-scheme: dark)`

assets/css/includes/scroll-bar.css
```diff
@@ -3,17 +3,13 @@
     background: 0 0;
 }
 
-.list:not(.dark)::-webkit-scrollbar-track {
-    background: var(--code-bg);
-}
-
```
taoyouh: I don't quite follow the changes on scroll bars here, but I don't notice any difference in my testing so I assume these are valid changes.

assets/css/common/post-single.css
adityatelange: it's not same thing twice. `@media` is used only when user has js disabled.

assets/css/includes/scroll-bar.css
adityatelange: yes this was redundant it seems. or some reason I don't remember about, but it doesn't affect it now.

assets/css/includes/scroll-bar.css
```diff
@@ -3,17 +3,13 @@
...
 ::-webkit-scrollbar-thumb {
     background: var(--tertiary);
     border: 5px solid var(--theme);
     border-radius: var(--radius);
 }
 
-.list:not(.dark)::-webkit-scrollbar-thumb {
+[data-theme="light"] .list::-webkit-scrollbar-thumb {
```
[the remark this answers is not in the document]
adityatelange: again media is used only when user has js disabled. it's inside noscript tag.

assets/css/common/post-single.css
taoyouh: Yeah I know there were 2 cases, but with `light-dark` you can write it once and cover the 2 cases

taoyouh:
Looks good to me now

adityatelange (approved):

PR #1833, opened 2026-02-15
Allow certain pages to be excluded from RSS feed

gbeine:
This line adds support for the possibility to exclude a certain page from the RSS feed by adding a param to the front matter:

```
---
rss:
  exclude: true
---
```

This patch allows certain pages to be excluded from the RSS feed.

There are a lot of sites around that discuss solutions like this.
It seems to be a common issue.

## PR Checklist

- [-] This change adds/updates translations and I have used the [template present here](https://github.com/adityatelange/hugo-PaperMod/wiki/Translations#want-to-add-your-language-).
- [-] This change adds a Social Icon which has a permissive license to use it.
- [-] This change updates the overridden internal templates from HUGO's repository.

### DIFF
--- a/layouts/_default/rss.xml
+++ b/layouts/_default/rss.xml
@@ -38,6 +38,7 @@
 {{- else }}
 {{- $pages = $pctx.Pages }}
 {{- end }}
+{{- $pages = where $pages "Params.rss.exclude" "!=" true -}}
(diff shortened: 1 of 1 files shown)


adityatelange (approved):
Renamed it to `Params.hiddenInRss`

gbeine:
Thank you very much!

PR #1840, opened 2026-02-28
Fix incorrect language prefix in menu URLs

leemojiang:

This PR fixes an issue with incorrect language prefix placement in menu URLs when using a `baseURL` that includes a subpath (e.g., `/blog/`). Previously, when using multilingual configurations (e.g., `zh` and `en`), the `zh` language prefix was incorrectly inserted before the baseURL subpath, resulting in URLs like:
https://leemojiang.github.io/zh/blog/archives/

Instead of the expected:
https://leemojiang.github.io/blog/zh/archives/

What changed
Detect external links (http://, https://, protocol-relative) and skip Hugo internal URL rewriting for them.
For internal links:
Normalize menu paths by removing a leading /.
Remove the baseURL subpath prefix from menu entries when present (to avoid duplicated /blog).
Rebuild links using site.Home.RelPermalink + menu_path so language prefix and subpath order is always correct.
Keep active menu highlighting based on normalized RelPermalink comparison to avoid false mismatches.
This improves compatibility for multilingual deployments under subpaths (for example, GitHub Pages project sites).

You can see an example of the fixed behavior on my blog: [https://leemojiang.github.io/blog/](https://leemojiang.github.io/blog/)

---

Yes, this issue is related to [#1661](https://github.com/adityatelange/hugo-PaperMod/issues/1661), where users reported incorrect URL generation when using a `baseURL` with a subpath and multiple languages. A similar issue is also discussed in another theme [#682](https://github.com/CaiJimmy/hugo-theme-stack/issues/682).

---

## PR Checklist

- [x] This change does not include any CDN resources/links.
- [x] This change does not include any unrelated scripts such as bash and python scripts.

### DIFF
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -83,28 +83,41 @@
         {{- $currentPage := . }}
         <ul id="menu">
             {{- range site.Menus.main }}
-            {{- $menu_item_url := (cond (strings.HasSuffix .URL "/") .URL (printf "%s/" .URL) ) | absLangURL }}
-            {{- $page_url:= $currentPage.Permalink | absLangURL }}
-            {{- $is_search := eq (site.GetPage .KeyName).Layout `search` }}
-            <li>
-                <a href="{{ .URL | absLangURL }}" title="{{ .Title | default .Name }} {{- cond $is_search (" (Alt + /)" | safeHTMLAttr) ("" | safeHTMLAttr ) }}"
-                {{- cond $is_search (" accesskey=/" | safeHTMLAttr) ("" | safeHTMLAttr ) }}>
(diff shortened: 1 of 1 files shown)


leemojiang:
Hi, I’ve updated the branch and all checks have passed.
If you have time, could you please take a look? Thank you!  @adityatelange

adityatelange:
Hi, I dont see this issue with demo site. That probably would be some issue in the hugo version you were testing. Thanks

See, 

- https://adityatelange.github.io/hugo-PaperMod/fr/archives/

- https://github.com/adityatelange/hugo-PaperMod/blob/5f104d743bf53a1b28490f9939e946f6e986be02/config.yml#L3

- https://github.com/adityatelange/hugo-PaperMod/blob/exampleSite/content/archives.fr.md?plain=1

PR #1842, opened 2026-03-03
fix: inherit case for tags from frontmatter

timonrieger:

This PR updates the single post tag rendering to use taxonomy term `.Name` instead of `.LinkTitle` in `layouts/_default/single.html`.

Using `.LinkTitle` causes tags to be title-cased by default in many setups (e.g. `programming` -> `Programming`) even when the source taxonomy term is lowercase. Switching to `.Name` makes tag labels in post footers match taxonomy/terms pages and preserves the original term casing from content/front matter.

Not yet.  
I can open an issue/discussion if maintainers prefer review there first.

## PR Checklist

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -46,7 +46,7 @@
     {{- $tags := .Language.Params.Taxonomies.tag | default "tags" }}
     <ul class="post-tags">
       {{- range ($.GetTerms $tags) }}
-      <li><a href="{{ .Permalink }}">{{ .LinkTitle }}</a></li>
+      <li><a href="{{ .Permalink }}">{{ .Name }}</a></li>
       {{- end }}
     </ul>
     {{- if (.Param "ShowPostNavLinks") }}


adityatelange:
Hi, this is intended, looks consistent across the blog. Users tend to put any casing, sometimes upper sometimes lower, which look inconsistent. Thanks

timonrieger:
hmm, not sure if I am getting your point. if you want to normalize casing of tags, it should apply globally, not only in some views. consider your demo site. It title-cases the tags on the single page, but does not alter it on the tag page. My patch would fix the first view, to not alter tag casing (as already done in the tag page)
- https://github.com/adityatelange/hugo-PaperMod/blob/exampleSite/content/posts/rich-content.md?plain=1
image

- https://github.com/adityatelange/hugo-PaperMod/blob/exampleSite/content/posts/papermod/papermod-faq.md?plain=1
image
image

timonrieger:
@adityatelange are you reconsidering this?

PR #1846, opened 2026-03-15
feat: add extend_post_content.html hook partials for post sections

### LINKED ISSUE #1845: [Enhancement]: partial injection hooks in single.html to avoid full layout overrides
## Problem

I need to inject custom content into `single.html` (e.g., after post content, before the footer), the only option that I have is to copy the entire `layouts/_default/single.html` into their project and maintain it as an override.

This creates a maintenance burden: every time PaperMod updates `single.html`, I have to manually diff and re-apply their customizations — or risk falling behind on upstream improvements.

## Proposed Solution

Add no-op hook partials at key injection points in `single.html` that users can override in their own `layouts/partials/` without touching the base layout.

**Example hooks in `single.html`:**
```html
{{/* Hook: users override this partial to inject content after post body */}}
{{- partial "extend_post_content.html" . }}
```

Ship them as empty stubs in PaperMod:
```
layouts/partials/extend_post_content.html  → empty
```

## Real-World Use Case

I maintain a quiz and flashcard system. I inject two partials after the post body, but must override the full `single.html` to do so. With an `extend_post_content.html` hook, I could drop the layout override entirely.

## Suggested Hook Locations

| Hook partial | Position |
|---|---|
| `layouts/partials/extend_post_content.html` | After content, before footer |

## Impact

Zero breaking changes — existing sites unaffected. Eliminates the most common reason users override `single.html`

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

henrypham67:

Adds an `extend_post_content.html` hook partial at a key injection point in `single.html`, after the post body and before the footer.

Currently, injecting custom content into `single.html` (e.g., quizzes, flashcard widgets, custom sections after post body) requires copying the entire `layouts/_default/single.html` into the project as a full override. This creates an ongoing maintenance burden — every upstream change to `single.html` requires a manual diff and re-merge.

This PR adds a single no-op stub partial that users can override in their own `layouts/partials/` without touching the base layout:

```html
{{- partial "extend_post_content.html" . }}
```

The stub ships empty (comments only), so existing sites are completely unaffected. This follows the same pattern already established by `extend_head.html` and `extend_footer.html` in PaperMod.

Yes. Closes #1845

## PR Checklist

---

**Commit quality assessment:** The commit is solid — minimal, focused, zero-breaking-changes, and consistent with PaperMod's existing `extend_head.html` / `extend_footer.html` pattern. The stub file comments clearly explain the purpose and override mechanism. Good to go.


adityatelange:
use comments.html?

henrypham67:
Thanks for the suggestion! I considered using `comments.html` as a hook, but it felt like the wrong trade-off — PaperMod only renders it when `comments: true` is set in frontmatter, which would be semantically misleading for my use case (flashcards/quiz content). It'd also create a collision point if real comment support (Disqus, utterances, etc.) is added later.
The current `single.html` override is minimal (two `if` blocks) and follows standard Hugo layout extension patterns. Keeping the concern in its own partial is cleaner than repurposing a comments hook for unrelated UI. Happy to discuss further if you see a downside I'm missing!

adityatelange:
it is not about how minimal the change is but about how many extend templates we can add for each template which user is not wiling to override for their own convenience.

PR #1849, opened 2026-03-30
Support Markdown in post description

lehmannro:

I'd like to use the `description` parameter in post frontmatter (particularly to show some content before the cover image). However, that content currently can only be plain text.  I'd like for them to be fed through `| markdownify` so I can at least use inline markup (`**bold**`, `*italics*`, etc).

I didn't want to change existing usage, so I'm introducing a `markdownifyDescription` parameter here that enables this new behavior.

I tried applying this not just to `single.html` but all places where `.Description` is used.

**Was the change discussed in an issue or in the Discussions before?**  Not that I know of.

## PR Checklist

### DIFF
--- a/layouts/_default/archives.html
+++ b/layouts/_default/archives.html
@@ -22,7 +22,11 @@
   </h1>
   {{- if .Description }}
   <div class="post-description">
-    {{ .Description }}
+    {{- if (.Param "markdownifyDescription") }}
+      {{ .Description | markdownify }}
+    {{- else }}
+      {{ .Description }}
+    {{- end }}
   </div>
   {{- end }}
 </header>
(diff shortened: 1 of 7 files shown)


adityatelange:
Hi, I don't think its a good idea for description to have other added content to be added before cover image. Please overide the particular template for adding a new field which is not a description of the post.

lehmannro:
Hey Aditya, thanks for your feedback!  Not sure I full understand, are you asking me to introduce a new parameter other than `description` (say, `preCoverLead`) that allows doing that?  [edit: `description` already adds content before the cover, so not sure I understand the feedback at all.]

Or are you against that type of lead-in at all?  If so, is there a way for me to override this template that's easier than maintaining, in perpetuity, a fork of a 70-line file (for a 1-line change)?  Maybe we could move that part to a smaller template I can override?

lehmannro:
Actually, no, on further thought:  I believe this PR shows there's **at least** a subtle bug in `layouts/partials/templates/schema_json.html`:  That file applies `plainify` to `Description`, which *none of the others* do.

